PR #16, opened 2025-10-23
Release/1.0.0

ffmcgee725:
## Explanation

Initial release for
- `@metamask/connect`
- `@metamask/connect-multichain`
- `@metamask/multichain-ui`
- `@metamask/analytics`

## Checklist


packages/connect/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/connect",
-  "version": "0.0.0",
+  "version": "0.1.0",
```
adonesky1, line 3: Is the exports section of this file correct?

ffmcgee725: Seems to make sense to me, is there something making you 🤨 ?

ffmcgee725: Actually, I believe it should be 

```json
"exports": {
    ".": {
      "browser": {
        "types": "./dist/connect-multichain/index.d.ts",
        "import": "./dist/connect-multichain/index.mjs"
      },
      "react-native": {
        "types": "./dist/connect-multichain/index.d.ts",
        "import": "./dist/connect-multichain/index.mjs"
      },
      "import": {
        "types": "./dist/connect-multichain/index.d.ts",
        "default": "./dist/connect-multichain/index.mjs"
      },
      "require": {
        "types": "./dist/connect-multichain/index.d.ts",
        "default": "./dist/connect-multichain/index.js"
      },
      "default": "./dist/connect-multichain/index.js"
    },
    "./package.json": "./package.json"
  },
```

no ? Notice, `connect-multichain` instead of `multichain`. This must have been forgotten in the [previous refactor](https://github.com/MetaMask/connect-monorepo/pull/14)

Another reminder is that we had agreed previously on exporting `connect-multichai` as the default experience, and then sub-routing `evm`, `bitcoin`, `legacy`, etc... onto the respective sub-routes

adonesky1: yeah I was meaning that 1) the package is renamed and 2) yeah it doesn't matter too much yet, but I'm not clear on how we're bundling/re-exporting

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@metamask/metamask-connect-monorepo",
-  "version": "0.0.0",
+  "version": "1.0.0",
```
adonesky1, line 3: [nit] should this just be 0.1.0 too?

packages/connect/package.json
jiexi: It's still `src/multichain/index.ts`, so the path in this package.json is correct. Purposely left it this way that way fuzzy file navigation isn't as annoying

jiexi: > not clear on how we're bundling/re-exporting

it gets compiled to just `export ... from '@metamask/connect-multichain'`. No bundling

package.json
jiexi: `create-release-branch` set this to `1.0.0`

jiexi:
I don't think having `"workspace:^"` as the dependency version will work after the package is published

ffmcgee725:
@jiexi actually you raise a good point.

Having `"workspace:^"` is not actually an issue, as at publish time, Yarn replaces `"workspace:^"` with a normal caret range of the referenced workspace’s version in the packed/published manifest, meaning for example, `"@metamask/connect-multichain": "workspace:^"` → published as `"@metamask/connect-multichain": "^0.1.0"` is `@metamask/connect-multichain` is already published as such.

Our issue, is that currently none of the packages are actually published to NPM, meaning, to avoid any window where installs fail, we should publish dependencies first, then dependents.

Now, looking at how https://github.com/MetaMask/action-npm-publish is setup, this action currently does not publish in topological order.

So I believe the solution would be to enforce topological order in our workflow by running a Yarn foreach publish step ourselves.

package.json
mcmire: Yes exactly. The way that we have `core` and other monorepos set up is that each new monorepo release bumps the root package's version by a major version. The release workflow will also create Git tags which points to the release commit, one of which is named after the root package (so in this case it would create `v1.0.0`). Finally it will also create a GitHub release and the title of that release will be the root version.

packages/analytics/CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
-### Changed
+## [0.1.0]
 
-- Bump `typescript` from ~5.2.2 to ~5.9.2 ([#11](https://github.com/MetaMask/connect-monorepo/pull/11))
+### Added
 
-[Unreleased]: https://github.com/MetaMask/metamask-connect-monorepo/
+- Initial release
```
mcmire: Nit: Ideally each changelog entry should have an associated PR. Perhaps we could  link this entry to this very release PR, e.g.

``` suggestion
- Initial release ([#16](https://github.com/MetaMask/connect-monorepo/pull/16))
```

(Same for the other packages. I haven't added suggestions to them but the same thing applies.)

ffmcgee725:
@jiexi after discussion with @mcmire ( https://consensys.slack.com/archives/C09AM4DJAUR/p1761314333781649?thread_ts=1761213978.773699&cid=C09AM4DJAUR ), I learned that as long as the dependencies refer to real packages locally then Yarn will be happy. and it doesn't matter which order packages are published in — they will all be published by the time the release workflow completes, so Yarn will do that replacement automatically when you run `yarn npm publish`

source: https://yarnpkg.com/features/workspaces#cross-references

So I'm reverting https://github.com/MetaMask/connect-monorepo/pull/16/commits/237cfe352fd78d2e213c28294de1614e93f822e1 for simplicity

adonesky1 (approved):

PR #21, opened 2025-10-23
feat: Legacy EVM Wrapper

wenfix:
## Usage
Test out the changes by running these at repo root:
1. `yarn && yarn build` 
2. `yarn workspace legacy-evm-react-vite dev`

## Checklist


packages/connect-evm/src/utils/get-eth-accounts.ts
```diff
@@ -0,0 +1,30 @@
+import type { SessionData } from '@metamask/connect-multichain';
+
+import type { Address } from '../types';
+
+/**
+ * TODO: Replace with getEthAccounts from @metamask/chain-agnostic-permission once Buffer dependency is removed
```
adonesky1: @wenfix can you remind me the issue here and start a thread with @mcmire since when I went to articulate the issue originally I couldn't find the words

packages/connect-evm/src/utils/type-guards.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @returns True if the request is a connect-like request, false otherwise
+ */
+export function isConnectRequest(req: ProviderRequest): req is Extract<
+  ProviderRequest,
+  {
+    method:
+      | 'wallet_requestPermissions'
+      | 'eth_requestAccounts'
+      | 'eth_accounts'
+      | 'eth_coinbase';
```
adonesky1:
```suggestion
```

🤔 I don't believe `eth_coinbase` is a connection request method. I just tested and it doesn't prompt a connection request?

packages/connect-evm/src/utils/type-guards.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+> {
+  return (
+    req.method === 'wallet_requestPermissions' ||
+    req.method === 'eth_requestAccounts' ||
+    req.method === 'eth_accounts' ||
+    req.method === 'eth_coinbase'
```
adonesky1:
```suggestion
    req.method === 'eth_accounts'
```

packages/connect-evm/src/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+  'eth_requestAccounts',
+  'eth_accounts',
+  'eth_coinbase',
+];
+
+export const DISTINCT_METHODS = [
+  'wallet_revokePermissions',
+  'wallet_switchEthereumChain',
+  'wallet_addEthereumChain',
+];
```
adonesky1: Not sure I follow the naming logic of `IGNORED_METHODS` and `DISTINCT_METHODS`

Why are they distinct?

adonesky1: [nit] maybe just included them directly in the `INTERCEPTABLE_METHODS` array?

adonesky1: I guess `IGNORED_METHODS` makes sense

packages/connect-evm/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@metamask/connect-evm",
+  "version": "0.1.0",
+  "description": "EVM Layer for MetaMask Connect",
```
adonesky1, line 4:
```suggestion
  "description": "EVM-specific interface for MetaMask Connect",
```

packages/connect-evm/README.md
```diff
@@ -0,0 +1,14 @@
+# `@metamask/connect-evm`
+
+EVM Layer for Metamask SDK
```
adonesky1: Maybe?
```suggestion
EVM-specific interface for MetaMask Connect
```

packages/connect-evm/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Installation
+
+`yarn add @metamask/connect/evm`
+
+### Development
+
```
adonesky1, line 127: We probably want to add some basic examples of how to consume this package

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+        data: {
+          name: 'metamask-provider',
+          data: request,
+        },
+      },
+      // eslint-disable-next-line no-restricted-globals
+      location.origin,
+    );
+  }
+}
```
adonesky1: I'm still confused, don't we need this to conditionally use other transports (MWP for instance) depending on the environment?

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+     * Handles the wallet_sessionChanged event.
+     * Updates the internal connection state with the new session data.
+     *
+     * @param session - The session data
+     */
+    this.#sessionChangedHandler = (session): void => {
+      this.#sessionScopes = session?.sessionScopes ?? {};
+
+      const ethAccounts = getEthAccounts(this.#sessionScopes);
+      this.#currentAccount = ethAccounts[0];
```
adonesky1: I don't believe this right. You won't be able to get the selected account from the sessionScopes. The only way to determine whether its changed is by listening to the `accountsChanged` event

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+    chainConfiguration,
+  }: {
+    chainId: number;
+    chainConfiguration?: AddEthereumChainParameter;
+  }): void {
+    if (this.#currentChainId === chainId) {
+      return;
+    }
+
+    // TODO: Check if approved scopes have the chain and early return
```
adonesky1, line 303: 👍

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    // Save the chain configuration for adding in case
+    // the chain is not configured in the wallet.
+    this.#latestChainConfiguration = chainConfiguration;
+
+    this.#request({
```
adonesky1: Shouldn't this be using the public `this.request` so it filters into the requestInterceptor?

packages/connect-evm/src/utils/get-eth-accounts.ts
```diff
@@ -0,0 +1,30 @@
...
+  const accounts: Address[] = [];
+  Object.entries(sessionScopes).forEach(([scope, scopeData]) => {
+    if (scope.startsWith('eip155:')) {
+      scopeData.accounts?.forEach((account) => {
+        const address = account.split(':').slice(2).join(':') as Address;
+        accounts.push(address);
+      });
+    }
+  });
+  return accounts;
```
jiexi: need to unique this

packages/connect-evm/src/utils/get-eth-accounts.ts
```diff
@@ -0,0 +1,30 @@
...
+): Address[] => {
+  if (!sessionScopes) {
+    return [];
+  }
+
+  const accounts: Address[] = [];
+  Object.entries(sessionScopes).forEach(([scope, scopeData]) => {
+    if (scope.startsWith('eip155:')) {
+      scopeData.accounts?.forEach((account) => {
+        const address = account.split(':').slice(2).join(':') as Address;
```
jiexi:
```suggestion
        const [_scope, _chain, address as Address] = account.split(':');
```
thoughts on something like this? Is it possible to use `parseCaipAccountId()` instead?

packages/connect-evm/src/utils/type-guards.ts
jiexi: `eth_accounts` isn't either then, right?

packages/connect-evm/src/utils/type-guards.ts
jiexi: if this list gets longer, it'd generally be better to define a constant and then use `includes()`

packages/connect-evm/src/connect.test.ts
```diff
@@ -0,0 +1,7 @@
+import { describe, it, expect } from 'vitest';
```
jiexi, line 1: we're currently doing `import * as t from 'vitest';` everywhere else in the codebase, but I prefer what you're doing here

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+
+/**
+ * Converts a number or string to a hex string
+ *
+ * @param value - The value to convert to hex
+ * @returns The hex value
+ */
+function toHex(value: number | string): string {
+  return `0x${value.toString(16)}`;
+}
```
jiexi: `@metamask/utils` has a `toHex` helper i believe

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+  readonly #core: MultichainCore;
+
+  /** An instance of the EIP-1193 provider interface */
+  readonly #provider: EIP1193Provider;
+
+  /** The currently selected chain ID on the wallet */
+  #currentChainId?: number | undefined;
+
+  /** The currently selected account on the wallet */
+  #currentAccount?: Address | undefined;
```
jiexi: thoughts on making this a getter that reads from `accounts`?

```
get currentAccount(): Address | undefined { 
  return accounts[0]  
}
```

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+        if (data?.method === 'metamask_chainChanged') {
+          const chainId = Number(data?.params.chainId);
+          this.#currentChainId = chainId;
+
+          // TODO: (@wenfix): better setter?
+          this.#provider.currentChainId = chainId;
+          this.#provider.emit('chainChanged', { chainId });
+          this.#eventHandlers?.chainChanged?.(chainId.toString());
```
jiexi: yeah might make sense to group these together in a setter

packages/connect-evm/src/connect.ts
jiexi: What Alex said is correct

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+    if (request.method === 'wallet_revokePermissions') {
+      return this.disconnect();
+    }
+
+    if (isConnectRequest(request)) {
+      return this.connect({
+        chainId: request.params[0] ?? 1,
+        account: request.params[1],
+      });
+    }
```
jiexi, line 360: don't fully understand how this block is functioning given the very different shapes that `eth_requestAccounts`  and `wallet_requestPermissions` have

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+      return this.switchChain({
+        chainId: parseInt(request.params[0].chainId, 16),
+      });
+    }
+
+    if (isAddChainRequest(request)) {
+      return this.#addEthereumChain(request.params[0]);
+    }
+
+    return Promise.resolve();
```
jiexi, line 377: above we resolve undefined explicitly. here it is implicitly. and then we have various other values. Should #requestInterceptor return `Promise<void>` as the type?

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+   * @param chainConfiguration - The chain configuration to use in case the chain is not present by the wallet
+   */
+  #addEthereumChain(chainConfiguration?: AddEthereumChainParameter): void {
+    if (!chainConfiguration && !this.#latestChainConfiguration) {
+      throw new Error('No chain configuration found.');
+    }
+
+    this.#request({
+      method: 'wallet_addEthereumChain',
+      params: [this.#latestChainConfiguration],
```
jiexi: should this be `chainConfiguration ?? this.#latestChainConfiguration`?

packages/connect-evm/src/connect.ts
jiexi: definitely still do. This only works for Extension (Firefox and Chromium), Mobile IAB. We need other logic to handle QR code from desktop/nodejs, and Mobile native browser environments

packages/connect-evm/src/provider.ts
```diff
@@ -0,0 +1,55 @@
...
+import { EventEmitter } from '@metamask/connect-multichain';
+
+import { INTERCEPTABLE_METHODS } from './constants';
+import type {
+  EIP1193ProviderEvents,
+  ProviderRequest,
+  ProviderRequestInterceptor,
+} from './types';
+
+export const EIP155 = 'eip155';
```
jiexi: i don't think this buys us too much more clarity haha 😅

packages/connect-evm/src/provider.ts
```diff
@@ -0,0 +1,55 @@
...
+   * Performs a EIP-1193 request.
+   *
+   * @param request - The request object containing the method and params
+   * @returns The result of the request
+   */
+  async request(request: ProviderRequest): Promise<unknown> {
+    /* Some methods require special handling, so we intercept them here
+     * and handle them in MetamaskConnectEVM.requestInterceptor method.  */
+    if (INTERCEPTABLE_METHODS.includes(request.method)) {
+      return this.#requestInterceptor?.(request);
```
jiexi, line 51: We have checks here for the method and in `#requestInterceptor`. Thoughts on having requestInterceptor return a boolean, true if request was handled, false if it should be handled here and then getting rid of the guard here?

packages/connect-evm/src/provider.ts
```diff
@@ -0,0 +1,55 @@
...
+  ProviderRequest,
+  ProviderRequestInterceptor,
+} from './types';
+
+export const EIP155 = 'eip155';
+
+/**
+ * EIP-1193 Provider wrapper around the Multichain SDK.
+ */
+export class EIP1193Provider extends EventEmitter<EIP1193ProviderEvents> {
```
jiexi, line 18: is this meant to be a full ~~EIP-1193~~ MetaMask window.ethereum equivalent provider? It seems this is technically EIP-1193 provider complaint, but our injected provider exposes other properties like `chainId` and `selectedAddress` on the provider object

packages/connect-evm/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+
+// Specific provider request types
+type ConnectRequest = {
+  method:
+    | 'wallet_requestPermissions'
+    | 'eth_requestAccounts'
+    | 'eth_accounts'
+    | 'eth_coinbase';
+  params: [chainId?: number, account?: string];
+  chainId?: string;
```
jiexi: not fully understanding this type

packages/connect-evm/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+type SwitchEthereumChainRequest = {
+  method: 'wallet_switchEthereumChain';
+  params: [{ chainId: string }];
+  chainId?: string;
+};
+
+type AddEthereumChainRequest = {
+  method: 'wallet_addEthereumChain';
+  params: [AddEthereumChainParameter];
+  chainId?: string;
```
jiexi: why do these all have optional `chainId` properties?

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,360 @@
...
+    this.#sessionChangedHandler = (session): void => {
+      this.#sessionScopes = session?.sessionScopes ?? {};
+
+      const ethAccounts = getEthAccounts(this.#sessionScopes);
+      this.#currentAccount = ethAccounts[0];
+      this.accounts = ethAccounts;
+    };
+
+    // eslint-disable-next-line no-restricted-globals
+    window.addEventListener('message', this.#metamaskProviderHandler);
```
jiexi: multiple sources of truth is awkward. Just wondering if there is a specific case that led you to put this lister in here as opposed to somewhere earlier in the flow

packages/connect-evm/src/utils/type-guards.ts
adonesky1: True

packages/connect-evm/src/connect.test.ts
wenfix: Good catch. Any benefit of doing the `import *`? Don't wanna change the style here but also prefer this. (Not that it's doing anything tho 😅 )

packages/connect-evm/src/connect.ts
wenfix: Its the other way around, the interceptor calls this function if it receives a `walllet_switchEthereumChain` request. If that fails, its handled by the event listener passed in this class' constructor and it calls `this.#addEthereumChain()`.

packages/connect-evm/src/utils/get-eth-accounts.ts
wenfix: 6f1f17e4

packages/connect-evm/src/utils/type-guards.ts
wenfix: 865c79bf

packages/connect-evm/src/utils/type-guards.ts
wenfix: Reduced by 865c79bf

packages/connect-evm/src/connect.ts
wenfix: cf082ec8

packages/connect-evm/src/connect.ts
wenfix: 50d532b7

packages/connect-evm/src/connect.ts
wenfix: I don't think we can because the sdk's `invokeMethod()` is expected to return the request's result (if applicable), so ideally this interceptors should aim to return a value. Does this make sense?

packages/connect-evm/src/connect.ts
wenfix: 946d916c

packages/connect-evm/src/provider.ts
wenfix: That's how I did it originally but felt kinda dirty, especially given that some methods return values (those that go through connect). So some would have returns which we could infer as truthy and some we'd have to explicitly return true. Was also concerned that it could lead to errors if a falsy value is returned from a method, as it would fallback to sending it through the Multichain API. Am I overthinking this? 😅

packages/connect-evm/src/types.ts
wenfix: Initially thought it made sense to be able to override the `currentSelectedChain` through the request but in retrospective it really doesn't make any sense 🤦

packages/connect-evm/src/types.ts
wenfix: https://github.com/MetaMask/connect-monorepo/pull/21/files#r2466917287

same as here ☝️

packages/connect-evm/src/connect.ts
jiexi: spaces vs tabs? doesn't matter because i can't read indents

jiexi: 🙃

packages/connect-evm/src/connect.ts
adonesky1: I see should switchChain be marked private too then (`#switchChain`) since its only supposed to be called internal to this class? Or will be called externally as well?

packages/connect-evm/src/constants.ts
adonesky1: I would still change `DISTINCT_METHODS` or just inline those methods in the `INTERCEPTABLE_METHODS` array since its not clear what the grouping criteria is

packages/connect-evm/package.json
```diff
@@ -0,0 +1,67 @@
...
+  },
+  "license": "MIT",
+  "sideEffects": false,
+  "exports": {
+    ".": {
+      "types": "./dist/browser/es/index.d.ts",
+      "import": "./dist/browser/es/connect-evm.mjs",
+      "default": "./dist/browser/es/connect-evm.mjs"
+    },
+    "./package.json": "./package.json"
```
ffmcgee725, line 25: We need to update this to account for `react-native` and `node` environments, same as `connect-multichain`

packages/connect-evm/src/connect.test.ts
ffmcgee725: I don't think we need to stick to `import * as t from 'vitest';` pattern here, might as well move away from this pattern (which I just left because migrating code, not actively refactoring it while doing so)

packages/connect-evm/src/connect.ts
ffmcgee725: ☝🏾 ➕

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,367 @@
...
+    if (isConnectRequest(request)) {
+      return this.connect({
+        chainId: request.params[0] ?? 1,
+        account: request.params[1],
+      });
+    }
+
+    if (isSwitchChainRequest(request)) {
+      return this.switchChain({
+        chainId: parseInt(request.params[0].chainId, 16),
```
ffmcgee725, line 364: do we have a `fromHex` function in `@metamask/utils` we could use here ?

packages/connect-evm/src/connect.ts
```diff
@@ -0,0 +1,367 @@
...
+    // eslint-disable-next-line no-restricted-globals
+    window.postMessage(
+      {
+        target: 'metamask-contentscript',
+        data: {
+          name: 'metamask-provider',
+          data: request,
+        },
+      },
+      // eslint-disable-next-line no-restricted-globals
```
ffmcgee725: if we end up needing these afterwards, we might as well add this to `eslint.config.mjs` instead of using the comments

packages/connect-evm/src/connect.ts
wenfix: I left it as public because the wagmi connector has a dedicated switchChain which could use this instead of writing the request manually. Happy to make it private if you think its more consistent!

packages/connect-evm/src/connect.ts
wenfix: 065011e1

packages/connect-evm/src/connect.ts
wenfix: 065011e1

packages/connect-evm/src/utils/get-eth-accounts.ts
wenfix: 939ad762

packages/connect-evm/src/connect.ts
wenfix: 939ad762

packages/connect-evm/src/constants.ts
wenfix: f334a3f6

packages/connect-evm/src/utils/get-eth-accounts.ts
wenfix: I added it here (939ad762) and it wasn't a problem. I'll need to investigate this a bit but for now it works with our test dapp. Will check if that's the case with wagmi.

packages/connect-evm/README.md
wenfix: a migration guide would also be sweet (maybe on another doc)

packages/connect-evm/src/provider.ts
wenfix: fa96c8c1

packages/connect-evm/package.json
jiexi: it's more so an EIP-1193 adapter/interface

packages/connect-evm/src/provider.ts
jiexi: fair, the return would be overloaded. Thank you for explaining

packages/connect-evm/package.json
jiexi: we might only need cjs and mjs exports? Because there's nothing directly in the `connect-evm` package itself that switches on the environment. That all lives in `connect-multichain`

I'm not 100% sure though

packages/connect-evm/src/connect.ts
jiexi: nit pick, i vote #private so we don't get devs calling it this way and then complaints later if we break it

jiexi:
@SocketSecurity ignore npm/@metamask/eth-json-rpc-infura@10.3.0

this can use fetch

jiexi:
@SocketSecurity ignore npm/builtin-status-codes@3.0.0

needs `http` for status codes

jiexi:
@SocketSecurity ignore npm/https-browserify@1.0.0

polyfill for https, makes sense if it needs `http` too

jiexi:
@SocketSecurity ignore npm/vite@5.4.21 makes sense, it's a build tool

jiexi:
@SocketSecurity ignore npm/@metamask/eth-query@4.0.0

we know that mcmire guy

jiexi:
@SocketSecurity ignore npm/@spruceid/siwe-parser@2.1.0

was published two years ago. New author not really relevant

jiexi:
@SocketSecurity ignore npm/assert@2.1.0

was published two years ago. New author not really relevant

jiexi:
@SocketSecurity ignore npm/browser-resolve@2.0.0
@SocketSecurity ignore npm/browserify-zlib@0.2.0
@SocketSecurity ignore npm/console-browserify@1.2.0
@SocketSecurity ignore npm/crypto-browserify@3.12.1
@SocketSecurity ignore npm/browserify-rsa@4.1.1
@SocketSecurity ignore npm/hash-base@3.0.5
@SocketSecurity ignore npm/md5.js@1.3.5
@SocketSecurity ignore npm/os-browserify@0.3.0
@SocketSecurity ignore npm/tty-browserify@0.0.1

popular package with version published awhile ago

jiexi:
@SocketSecurity ignore npm/ripemd160@2.0.3
@SocketSecurity ignore npm/sha.js@2.4.12

new author contributes/maintains other popular packages

adonesky1 (approved):
LGTM! I think we're stable enough now to take this to the bank and re-direct future changes to `main`

PR #31, opened 2025-11-06
feat: wagmi connector 

wenfix:
## Explanation

Adds Metamask's wagmi connector, wagmi's playground and several fixes and enhancements to ConnectEVM, listed below:

### EventEmitter Enhancement
- Added removeListener() to the internal EventEmitter class in connect-multichain

### Chain Switching Improvements
- Fixed ethereum_switchChain fallback when chain is not configured
- Automatically attempts to add the chain via wallet_addEthereumChain if switching fails due to unrecognized chain
- Is now synchronous and not dependent on listening to error notification from the wallet

 ### Permission Handling
- Fixed permissioning accounts through wallet_requestPermissions
- Added forceRequest flag to ensure connection prompts are shown when needed
- Handles cases where the SDK would skip connection requests for already-connected sessions

### Developer Experience
- Fixed Typedoc generation issues
- Added conditional logging through debug option in createMetamaskConnectEVM()
- Enables debug logging only when explicitly enabled via options
- Fix `react-multichain-playground` build by suppressing dependencies' source map warnings 

### For wagmi's playground

`yarn install && yarn build && cd integrations/wagmi && yarn start`

### For portofolio dogfooding
1. Clone [wagmi](https://github.com/wevm/wagmi/)
2. Replace `metamask.ts` w/ `metamask-connector.ts` from this branch
2. Add `@metamask/connect-evm` dependency to wagmi's `packages/connectors`
3. Publish both (can use [verdaccio](https://verdaccio.org/)) `@metamask/connect-evm` and `wagmi/connectors`(or local link it) to be consumed in portfolio.
4. Clone MetaMask Portfolio and add the deployed version of `wagmi/connectors` to the package's resolutions field.

## Checklist

### DIFF
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -136,9 +136,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       }
       // Fallback to requesting chainId from provider if SDK doesn't return it
       const provider = await this.getProvider();
-      if (!provider) {
-        throw new ProviderNotFoundError();
-      }
+
       const hexChainId = await provider.request({ method: 'eth_chainId' });
       return Number(hexChainId);
     },
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -1,53 +1,32 @@
 import {
   createMetamaskConnectEVM,
-  MetamaskConnectEVMOptions,
   type MetamaskConnectEVM,
 } from '@metamask/connect-evm';
-import { MultichainCore, SessionData } from '@metamask/connect-multichain';
 
-// import type {
-//   MultichainCore,
-//   SDKState,
-//   SessionData,
-// } from '@metamask/multichain-sdk';
-// import type { MetaMaskSDKOptions } from '@metamask/sdk';
+//import type { SDKState, SessionData } from '@metamask/multichain-sdk'
+import type { MetaMaskSDKOptions } from '@metamask/sdk';
 
 import {
   ChainNotConfiguredError,
-  type Connector,
+  //type Connector,
   createConnector,
-  extractRpcUrls,
   ProviderNotFoundError,
 } from '@wagmi/core';
 import type {
   Compute,
   ExactPartial,
   OneOf,
-  RemoveUndefined,
   UnionCompute,
 } from '@wagmi/core/internal';
 import {
-  type AddEthereumChainParameter,
-  type Address,
   type EIP1193Provider,
   getAddress,
-  type Hex,
-  hexToNumber,
-  numberToHex,
   type ProviderConnectInfo,
-  type ProviderRpcError,
-  ResourceUnavailableRpcError,
-  type RpcError,
   SwitchChainError,
-  UserRejectedRequestError,
-  withRetry,
-  withTimeout,
 } from 'viem';
 
-type MetaMaskSDK = MultichainCore;
-
 export type MetaMaskParameters = UnionCompute<
-  MetaMaskSDKOptions &
+  WagmiMetaMaskSDKOptions &
     OneOf<
       | {
           /* Shortcut to connect and sign a message */
@@ -61,30 +40,28 @@ export type MetaMaskParameters = UnionCompute<
     >
 >;
 
-// type WagmiMetaMaskSDKOptions = Compute<
-//   ExactPartial<
-//     Omit<
-//       MetaMaskSDKOptions,
-//       | '_source'
-//       | 'forceDeleteProvider'
-//       | 'forceInjectProvider'
-//       | 'injectProvider'
-//       | 'useDeeplink'
-//       | 'readonlyRPCMap'
-//     >
-//   > & {
-//     /** @deprecated */
-//     forceDeleteProvider?: MetaMaskSDKOptions['forceDeleteProvider'];
-//     /** @deprecated */
-//     forceInjectProvider?: MetaMaskSDKOptions['forceInjectProvider'];
-//     /** @deprecated */
-//     injectProvider?: MetaMaskSDKOptions['injectProvider'];
-//     /** @deprecated */
-//     useDeeplink?: MetaMaskSDKOptions['useDeeplink'];
-//   }
-// >;
-
-type MetaMaskSDKOptions = MetamaskConnectEVMOptions;
+type WagmiMetaMaskSDKOptions = Compute<
+  ExactPartial<
+    Omit<
+      MetaMaskSDKOptions,
+      | '_source'
+      | 'forceDeleteProvider'
+      | 'forceInjectProvider'
+      | 'injectProvider'
+      | 'useDeeplink'
+      | 'readonlyRPCMap'
+    >
+  > & {
+    /** @deprecated */
+    forceDeleteProvider?: MetaMaskSDKOptions['forceDeleteProvider'];
+    /** @deprecated */
+    forceInjectProvider?: MetaMaskSDKOptions['forceInjectProvider'];
+    /** @deprecated */
+    injectProvider?: MetaMaskSDKOptions['injectProvider'];
+    /** @deprecated */
+    useDeeplink?: MetaMaskSDKOptions['useDeeplink'];
+  }
+>;
 
 metaMask.type = 'metaMask' as const;
 export function metaMask(parameters: MetaMaskParameters = {}) {
@@ -93,18 +70,19 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
     onConnect(connectInfo: ProviderConnectInfo): void;
     onDisplayUri(uri: string): void;
   };
-  type Listener = Parameters<Provider['on']>[1];
+  //type Listener = Parameters<Provider['on']>[1]
 
   let metamask: MetamaskConnectEVM;
-  let provider: Provider | undefined;
-  let providerPromise: Promise<typeof provider>;
+  // let provider: Provider | undefined
+  // let providerPromise: Promise<typeof provider>
 
-  let accountsChanged: Connector['onAccountsChanged'] | undefined;
-  let chainChanged: Connector['onChainChanged'] | undefined;
-  let connect: Connector['onConnect'] | undefined;
-  let displayUri: ((uri: string) => void) | undefined;
-  let disconnect: Connector['onDisconnect'] | undefined;
-  let sessionData: SessionData | undefined;
+  // let accountsChanged: Connector['onAccountsChanged'] | undefined
+  // let chainChanged: Connector['onChainChanged'] | undefined
+  // let connect: Connector['onConnect'] | undefined
+  // let displayUri: ((uri: string) => void) | undefined
+  // let disconnect: Connector['onDisconnect'] | undefined
+  // let sessionData: SessionData | undefined
+  // let sdkState: SDKState | undefined
 
   return createConnector<Provider, Properties>((config) => ({
     id: 'metaMaskSDK',
@@ -121,13 +99,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         eventHandlers: {
           accountsChanged: this.onAccountsChanged.bind(this),
           chainChanged: this.onChainChanged.bind(this),
+          //@ts-expect-error cool
           connect: this.onConnect.bind(this),
           disconnect: this.onDisconnect.bind(this),
         },
       });
-
-      console.log('SDK created', metamask);
     },
+    //@ts-expect-error cool
     async connect({ chainId, isReconnecting, withCapabilities } = {}) {
       // TODO (@wenfix): handle case where no chainId is provided
       const _chainId = chainId ?? 1;
@@ -155,6 +133,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         chainId: result.chainId ?? _chainId,
       };
 
+      //@ts-expect-error cool
       config.emitter.emit('connect', response);
 
       return response;
@@ -177,6 +156,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       return 1;
     },
 
+    //@ts-expect-error cool
     async getProvider() {
       const provider = await metamask.getProvider();
       if (!provider) {
@@ -215,6 +195,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         addEthereumChainParameter,
       });
 
+      //@ts-expect-error cool
       metamask.switchChain({ chainId, chainConfiguration });
 
       return chain;
@@ -232,10 +213,11 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       config.emitter.emit('change', { chainId });
     },
     async onConnect(connectInfo) {
-      //handled internally
+      //@ts-expect-error cool
+      config.emitter.emit('connect', connectInfo);
     },
-    async onDisconnect(error) {
-      //handled internally
+    async onDisconnect() {
+      config.emitter.emit('disconnect');
     },
     async onDisplayUri(uri) {
       config.emitter.emit('message', { type: 'display_uri', data: uri });
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -229,7 +229,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       // https://github.com/MetaMask/providers/pull/120
       if (error && (error as unknown as RpcError<1013>).code === 1013) {
         const provider = await this.getProvider();
-        if (provider && !!(await this.getAccounts()).length) return;
+        if (provider && (await this.getAccounts()).length > 0) return;
       }
 
       config.emitter.emit('disconnect');
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -199,6 +199,8 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
     },
 
     async onAccountsChanged(accounts) {
+      // TODO: verify if this is needed or if we can just rely on the
+      // existing disconnect event instead
       // Disconnect if there are no accounts
       if (accounts.length === 0) {
         this.onDisconnect();
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -3,7 +3,6 @@ import {
   type MetamaskConnectEVM,
 } from '@metamask/connect-evm';
 
-//import type { SDKState, SessionData } from '@metamask/multichain-sdk'
 import type { MetaMaskSDKOptions } from '@metamask/sdk';
 
 import {
@@ -70,19 +69,8 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
     onConnect(connectInfo: ProviderConnectInfo): void;
     onDisplayUri(uri: string): void;
   };
-  //type Listener = Parameters<Provider['on']>[1]
 
   let metamask: MetamaskConnectEVM;
-  // let provider: Provider | undefined
-  // let providerPromise: Promise<typeof provider>
-
-  // let accountsChanged: Connector['onAccountsChanged'] | undefined
-  // let chainChanged: Connector['onChainChanged'] | undefined
-  // let connect: Connector['onConnect'] | undefined
-  // let displayUri: ((uri: string) => void) | undefined
-  // let disconnect: Connector['onDisconnect'] | undefined
-  // let sessionData: SessionData | undefined
-  // let sdkState: SDKState | undefined
 
   return createConnector<Provider, Properties>((config) => ({
     id: 'metaMaskSDK',
@@ -123,20 +111,17 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         account: undefined,
       } as { chainId: number; account: string });
 
+      // TODO: return this again after changing portfolio
+      // biome-ignore lint/correctness/noUnusedVariables: will be used in the future
       const accounts = result.accounts.map((account) => ({
         address: account,
         capabilities: {},
       }));
 
-      const response = {
-        accounts,
+      return {
+        accounts: result.accounts,
         chainId: result.chainId ?? _chainId,
       };
-
-      //@ts-expect-error cool
-      config.emitter.emit('connect', response);
-
-      return response;
     },
 
     async disconnect() {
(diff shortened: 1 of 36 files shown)


packages/connect-multichain/package.json
```diff
@@ -65,11 +65,11 @@
...
   },
   "dependencies": {
-    "@metamask/analytics": "workspace:^",
+    "@metamask/analytics": "^0.1.0",
     "@metamask/mobile-wallet-protocol-core": "^0.3.0",
     "@metamask/mobile-wallet-protocol-dapp-client": "^0.2.1",
-    "@metamask/multichain-api-client": "^0.8.1",
-    "@metamask/multichain-ui": "workspace:^",
+    "@metamask/multichain-api-client": "^0.8.0",
+    "@metamask/multichain-ui": "0.1.0",
```
jiexi: these changes need to be reverted i think?

packages/connect-evm/src/connect.ts
```diff
@@ -353,9 +346,17 @@ export class MetamaskConnectEVM {
...
+      // explicitly request a specific account. This is needed to workaround
+      // wallet_requestPermissions not requesting specific accounts.
+      const shouldForceConnectionRequest =
+        request.method === 'wallet_requestPermissions';
+
       return this.connect({
-        chainId: request.params[0] ?? 1,
-        account: request.params[1],
+        chainId: DEFAULT_CHAIN_ID,
+        forceRequest: shouldForceConnectionRequest,
```
jiexi, line 360: ahh... boy. Should this flow through this.connect() at all then? Maybe just don't intercept it?

jiexi: is it because we might not have accountsChanged and chainChanged listeners setup if we don't call connect().

Also I just noticed connect() doesn't use one time listeners, so each time we call connect() the listeners pile up

packages/connect-evm/src/connect.ts
```diff
@@ -132,24 +131,27 @@ export class MetamaskConnectEVM {
...
    * @param options.chainId - The chain ID to connect to (defaults to 1 for mainnet)
    * @param options.account - Optional specific account to connect to
+   * @param options.forceRequest - Wwhether to force a request regardless of an existing session
    * @returns A promise that resolves with the connected accounts and chain ID
    */
   async connect(
     {
       chainId,
       account,
+      forceRequest,
```
jiexi, line 139: don't love that we need this force flag. I feel like we can get around this by not intercepting `wallet_requestPermissions` and remapping it to `connect()`

integrations/wagmi/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react-jsx",
+
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true
+  },
+  "include": ["src", "./metamask-connector.ts"],
```
jiexi, line 20: is there a reason why `metamask-connector.ts` is not also nested under `src`?

integrations/wagmi/src/main.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client'
+import { Buffer } from 'buffer'
+import React from 'react'
+import ReactDOM from 'react-dom/client'
+import { deserialize, serialize, WagmiProvider } from 'wagmi'
+
+import './index.css'
+
+// `@coinbase-wallet/sdk` uses `Buffer`
+globalThis.Buffer = Buffer
```
jiexi, line 13: not the most robust way to polyfill, but this is fine for now

packages/connect-evm/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add wagmi connector and wagmi react playground ([#31](https://github.com/MetaMask/connect-monorepo/pull/31))
```
jiexi: should the metamask-connector be imported and exported in `connect-evm`? or should it be it's own package

packages/connect-evm/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add `removeListener` function to internal `EventEmitter` ([#31](https://github.com/MetaMask/connect-monorepo/pull/31))
```
jiexi: is there a user facing change here?

packages/connect-multichain/src/domain/events/index.ts
```diff
@@ -58,6 +58,21 @@ export class EventEmitter<TEvents extends Record<string, unknown[]>> {
...
+   * @template TEventName - The name of the event to remove the handler from (must be a key of TEvents)
+   * @param eventName - The name of the event to remove the handler from
+   * @param handler - The specific handler function to remove
+   */
+  removeListener<TEventName extends keyof TEvents & string>(
+    eventName: TEventName,
+    handler: (...eventArg: TEvents[TEventName]) => void,
+  ) {
+    this.#emitter.off(eventName, handler as (...args: any[]) => void);
+  }
```
jiexi, line 75: do we need to add addListener, once, etc?

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+      isReconnecting?: boolean | undefined;
+      withCapabilities?: withCapabilities | boolean | undefined;
+    }) {
+      // TODO: better handling when not providing a chainId?
+
+      const chainId = parameters?.chainId ?? 1;
+      const withCapabilities =
+        ('withCapabilities' in (parameters ?? {}) &&
+          parameters?.withCapabilities) ||
+        false;
```
jiexi: can this be simplified somehow? perhaps setting false as the default value in the params?

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+      if (accounts.length === 0) return;
+
+      const chainId = Number(connectInfo.chainId);
+      config.emitter.emit('connect', { accounts, chainId });
+    },
+
+    async onDisconnect(error?: RpcError) {
+      // TODO: (wenfix) Is this still necessary?
+
+      const provider = await this.getProvider();
```
jiexi: we only need to get the provider in the error code is 1013

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+      // If MetaMask emits a `code: 1013` error, wait for reconnection before disconnecting
+      // https://github.com/MetaMask/providers/pull/120
+      if (error && (error as unknown as RpcError<1013>).code === 1013) {
+        if (provider && !!(await this.getAccounts()).length) return;
```
jiexi: nit. can this be `.length > 0` rather than `!!`?

adonesky1:
naive attempt to test in the playground -> failing to connect
>For wagmi's playground
> `yarn install && yarn build && cd integrations/wagmi && yarn preview`

Screenshot 2025-11-20 at 3 21 41 PM

🤔 something weird is going on here. `readOnlyRpcMap` shouldn't be the name of the variable anymore. I ran `yarn clean` to clear caching... probably something off on my side


**EDIT**: needed to run `yarn vite build` in `integrations/wagmi` now seems to be working

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+    type: metaMask.type,
+    async setup() {
+      const supportedNetworks = Object.fromEntries(
+        config.chains.map((chain) => [
+          `eip155:${chain.id}`,
+          chain.rpcUrls.default?.http[0],
+        ]),
+      );
+
+      // TODO: check if we need to support other parameters
```
adonesky1, line 68: can you elaborate on which params you think we might still need?

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+          account: undefined,
+        });
+
+        return {
+          accounts: (withCapabilities
+            ? result.accounts.map((account) => ({
+                address: account,
+                capabilities: {},
+              }))
+            : result.accounts) as never,
```
adonesky1, line 105: oof painful typing, but looks right upon reviewing WAGMI docs

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+          chainId: result.chainId ?? chainId,
+        };
+      } catch (err) {
+        const error = err as RpcError;
+        if (error.code === UserRejectedRequestError.code)
+          throw new UserRejectedRequestError(error);
+        if (error.code === ResourceUnavailableRpcError.code)
+          throw new ResourceUnavailableRpcError(error);
+        throw error;
```
adonesky1: [nit] - I think we generally tend to use brackets in these cases
```suggestion
        if (error.code === UserRejectedRequestError.code) {
          throw new UserRejectedRequestError(error);
         }
        if (error.code === ResourceUnavailableRpcError.code) {
          throw new ResourceUnavailableRpcError(error);
         }
        throw error;
```

adonesky1:
Seeing a switch chain bug? 🔉 Sound on

https://github.com/user-attachments/assets/4041ed21-5e60-4370-ae11-5b40da515980

And more context:

https://github.com/user-attachments/assets/140899a1-34e4-482b-9686-e9ed2be9c43f

So we're not actually calling `wallet_switchEthereumChain` in the connect-evm layer since the chain is permitted. I know we went back and forth on this, but I'd thought we'd settled that we would actually call to the wallet when in the default transport but not when in the MWP transport? 

Screenshot 2025-11-20 at 4 33 35 PM

adonesky1:
Tons of chainChanged events. 🔉 sound on

https://github.com/user-attachments/assets/ed7e2999-6024-4976-adbc-520d7ea02220

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,248 @@
...
+        return Number(chainId);
+      }
+      // Fallback to requesting chainId from provider if SDK doesn't return it
+      const provider = await this.getProvider();
+      if (!provider) {
+        throw new ProviderNotFoundError();
+      }
+      const hexChainId = await provider.request({ method: 'eth_chainId' });
+      return Number(hexChainId);
+    },
```
adonesky1, line 138: Looks to me like we won't ever hit
```
if (!provider) {
        throw new ProviderNotFoundError();
      }
```
since we're doing the same thing in the `getProvider` method itself

packages/connect-evm/src/connect.ts
adonesky1: ➕

packages/connect-evm/CHANGELOG.md
adonesky1: Not sure if the intention was to export it and consume it in WAGMI directly, but I suppose that would make sense. It makes sense to me that it would live in `connect-evm` if we do intend to export/consume it like this - which now that I think about it we should do

integrations/wagmi/src/main.tsx
wenfix: this is just a port of wagmi's code as is, didn't make any changes and didn't plan to. One could even argue if it makes sense to keep this package in here.

integrations/wagmi/tsconfig.json
wenfix: Not being related to the playground code at all and being easier to reach. As I mention below, everything under `src` is just a port of wagmi's playground.

packages/connect-evm/src/connect.ts
wenfix: My reasoning for this is that, besides the logic that `connect` method handles in this layer, it delegates to the underlying `connect-multichain` `connect` method, which handles everything session and transport related. Doing it like that would essentially meaning that connection logic has 2 different code paths, one that goes through `connect` method and one that doesn't. I might be missing something but it doesn't seem like something we'd wanna do. Think we can omit the `forceRequest` parameter in the connect-evm and always force it, so we don't have to expose that.

packages/connect-evm/src/connect.ts
wenfix: see https://github.com/MetaMask/connect-monorepo/pull/31#discussion_r2549338381

packages/connect-evm/CHANGELOG.md
wenfix: We don't want to export the wagmi connector on our `connect-evm` package, as other consumers wouldn't need it. I also think it makes sense for the code to live in `wagmi`'s repo, as it only pertains to it, and their consumers would have to go through hoops to just look at the connector. The whole idea of bringing it here was to have a closer reference and a way to test it internally w/o the whole wagmi repo/setup.

packages/connect-evm/CHANGELOG.md
wenfix: Good catch!

packages/connect-multichain/src/domain/events/index.ts
wenfix: `once` might be useful but i've gone with a YAGNI approach on this. We can add it if needed?

packages/connect-evm/CHANGELOG.md
wenfix: But also who do we consider users here? The dev's that'll use it or actual end-users?

integrations/wagmi/metamask-connector.ts
wenfix: 26796cea

integrations/wagmi/metamask-connector.ts
wenfix: 26796cea

integrations/wagmi/metamask-connector.ts
wenfix: 02d5b745

integrations/wagmi/metamask-connector.ts
wenfix: no, can you? 👀 

jokes aside, this line has been here for a while because of the options listed in the [old sdk docs](https://docs.metamask.io/sdk/reference/sdk-options/). 

I'm now thinking that we'll learn which we need as we progress and have consumers and can add them later. thoughts?

integrations/wagmi/metamask-connector.ts
wenfix: 973777c9

integrations/wagmi/metamask-connector.ts
wenfix: yes?

acbcc0e4

wenfix:
> Seeing a switch chain bug? 🔉 Sound on
> 
>  Screen.Recording.2025-11-20.at.4.06.56.PM.mov 
> And more context:
> 
>  Screen.Recording.2025-11-20.at.4.08.08.PM.mov 
> So we're not actually calling `wallet_switchEthereumChain` in the connect-evm layer since the chain is permitted. I know we went back and forth on this, but I'd thought we'd settled that we would actually call to the wallet when in the default transport but not when in the MWP transport?
> 
> Screenshot 2025-11-20 at 4 33 35 PM

Yeah I think this makes sense, we just need to expose which type of transport we're using through the connect multichain I think 🤔

integrations/wagmi/metamask-connector.ts
adonesky1: Right I see. Which -  if any - were  in our original WAGMI connector that we've removed here?

packages/connect-multichain/package.json
wenfix: bb84bd70

integrations/wagmi/metamask-connector.ts
wenfix: you can see the options type [here](https://github.com/MetaMask/metamask-sdk/blob/338a07a7065c422fcb319c20b9f7d1489e484ce5/packages/sdk/src/sdk.ts#L38) w/ everything listed.

it was consumed in the metamask connector [here](https://github.com/wevm/wagmi/blob/0e0076bed87d314d96e8d1793eab29aa88eeadda/packages/connectors/src/metaMask.ts#L38)

packages/connect-evm/src/connect.ts
jiexi: > Think we can omit the forceRequest parameter in the connect-evm and always force it, so we don't have to expose that.

Yeah i think this would be preferable

integrations/wagmi/metamask-connector.ts
jiexi: >I'm now thinking that we'll learn which we need as we progress and have consumers and can add them later. thoughts?

i think this makes sense

packages/connect-evm/src/connect.ts
jiexi: replied to that comment

packages/connect-evm/CHANGELOG.md
jiexi: sorry, my question was more if adding a `removeListener` function to the `EventEmitter` was visible on the public interface of anything we export, or is it only seen internally. I should have checked the code myself, sorry about that.

So yes, this introduces the `removeListener` to `MultichainSDK` and the `EIP1193Provider`. I believe both the multichain and evm packages should get a changelog entry for this then. What are your thoughts?

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,242 @@
...
+  dapp?: CreateMetamaskConnectEVMParameters['dapp'] | undefined;
+} & OneOf<
+  | {
+      /* Shortcut to connect and sign a message */
+      connectAndSign?: string | undefined;
+    }
+  | {
+      // TODO: Strongly type `method` and `params`
+      /* Allow `connectWith` any rpc method */
+      connectWith?: { method: string; params: unknown[] } | undefined;
```
adonesky1, line 41: Not required as part of this PR but lets implement these on the WAGMI playground 🙏

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,242 @@
...
+        throw new SwitchChainError(error);
+      }
+    },
+
+    async onAccountsChanged(accounts) {
+      // Disconnect if there are no accounts
+      if (accounts.length === 0) {
+        this.onDisconnect();
+        return;
+      }
```
jiexi, line 208: i would think the existing `disconnect` event would cover this

```
disconnect: this.onDisconnect.bind(this),
```

Is this to be extra safe, or do we not see the `disconnect` event from the MMC instance firing reliably?

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,242 @@
...
+      // Fallback to requesting chainId from provider if SDK doesn't return it
+      const provider = await this.getProvider();
+
+      const hexChainId = await provider.request({ method: 'eth_chainId' });
+      return Number(hexChainId);
+    },
+
+    async getProvider() {
+      const provider = await metamask.getProvider();
```
jiexi, line 141: wait... why is `getProvider()` async in the first place? I don't see why it needs to be in ConnectEvm

integrations/wagmi/metamask-connector.ts
```diff
@@ -0,0 +1,242 @@
...
+      const chainId = Number(connectInfo.chainId);
+      config.emitter.emit('connect', { accounts, chainId });
+    },
+
+    async onDisconnect(error?: RpcError) {
+      // If MetaMask emits a `code: 1013` error, wait for reconnection before disconnecting
+      // https://github.com/MetaMask/providers/pull/120
+      if (error && (error as unknown as RpcError<1013>).code === 1013) {
+        const provider = await this.getProvider();
+        if (provider && (await this.getAccounts()).length > 0) return;
```
jiexi, line 234: not quite sure how this achieves waiting for a reconnection. Would doing a setTimeout make this more clear?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -131,6 +133,10 @@ export class MultichainSDK extends MultichainCore {
     return this.options.storage;
   }
 
+  get transportType(): TransportType {
+    return this.__transportType;
```
jiexi: Thoughts on determining this using `this.__transport instanceof ...` rather than having to store and keep this synchronized when we set `__transport`? Idk. not super important right now

integrations/wagmi/metamask-connector.ts
wenfix: c46ac65c

packages/connect-evm/CHANGELOG.md
wenfix: Your comment made me think that the `connect-evm` doesn't even need an entry for this, just `connect-multichain` as its where the methods are defined and implemented. `EIP1193Provider` extends our `EventEmitter` and that didn't change, does that make sense? I'm unsure if both entries make sense

packages/connect-multichain/src/multichain/index.ts
wenfix: I like it. c9cce41f

packages/connect-evm/CHANGELOG.md
jiexi: Right, but because `EIP1193Provider` extends the `EventEmtter` that did change, then itself has also changed. The current changelogs you have work for me

adonesky1:
Unable to get native browser working w/ WAGMI playground. Not seeing any logs that help debug, but digging a bit now

https://github.com/user-attachments/assets/f1240660-899f-4270-bd51-d6775a46c6d0

integrations/wagmi/metamask-connector.ts
jiexi: added comment here https://github.com/MetaMask/connect-monorepo/pull/31/commits/c1dadd6ec70fe3cca054c9396eefb80ba920d1d8

integrations/wagmi/metamask-connector.ts
jiexi: I'm not certain this is needed, but rather than adding a delay I'm going to just assume the subsequent `connect` event will put the provider back in the correct state

jiexi:
@SocketSecurity ignore-all

All these new deps are from the wagmi integration playground which is not published as a package

jiexi:
https://consensyssoftware.atlassian.net/browse/WAPI-868

packages/connect-evm/src/connect.ts
```diff
@@ -132,24 +129,27 @@ export class MetamaskConnectEVM {
    * @param options - The connection options
    * @param options.chainId - The chain ID to connect to (defaults to 1 for mainnet)
    * @param options.account - Optional specific account to connect to
+   * @param options.forceRequest - Wwhether to force a request regardless of an existing session
```
jiexi, line 132: typo. but not going to fix this now

packages/connect-evm/src/connect.ts
jiexi: https://consensyssoftware.atlassian.net/browse/WAPI-868

adonesky1 (approved):

PR #33, opened 2025-11-07
feat: add support for read only RPC calls

ffmcgee725:
## Explanation

The following PR :
1. Modifies EVM playground to include breakout buttons for specific read-only method (eth_getBalance, eth_blockNumber, and eth_gasPrice).
2. Adds `readonlyRpcMap` values to the `createConnectEvm` constructor in the EVM playground.
3. Ensures `readonlyRpcMap` is required for the EVM constructor.
4. Adds validation to enforce that only configured chains are callable.

## Checklist


packages/connect-evm/src/provider.ts
```diff
@@ -57,9 +57,41 @@ export class EIP1193Provider extends EventEmitter<EIP1193ProviderEvents> {
...
+      'eth_getCode',
+      'eth_call',
+      'eth_estimateGas',
+      'eth_getLogs',
+      'eth_getTransactionCount',
+      'eth_getBlockByNumber',
+      'eth_getBlockByHash',
+      'eth_getTransactionByHash',
+      'eth_getTransactionReceipt',
+    ].includes(request.method);
```
adonesky1: There is a list of readOnly methods in `connect-multichain/src/domain/multichain/api` called `RPC_HANDLED_METHODS` that we should be using here instead

packages/connect-evm/src/connect.ts
```diff
@@ -613,27 +615,43 @@ export class MetamaskConnectEVM {
...
+  options: Pick<MultichainOptions, 'dapp' | 'api'> & {
     eventEmitter?: MinimalEventEmitter;
     eventHandlers?: EventHandlers;
   },
 ): Promise<MetamaskConnectEVM> {
   logger('Creating Metamask Connect/EVM with options:', options);
 
+  // Validate that readOnlyRpcMap is provided and not empty
+  if (
+    !options.api?.readonlyRPCMap ||
```
jiexi, line 635: should we make this accept hex and then convert it to CAIP-2?

packages/connect-multichain/src/multichain/rpc/handlers/rpcClient.ts
```diff
@@ -25,21 +25,24 @@ export class RpcClient {
 	constructor(
 		private readonly config: MultichainOptions,
 		private readonly sdkInfo: string,
+		private readonly defaultTimeout: number = 30000, // 30 seconds default timeout
 	) {}
 
 	/**
 	 * Routes the request to a configured RPC node.
+	 * @param options - The invoke method options
+	 * @param options.timeout - Optional timeout in milliseconds (defaults to 30000ms)
```
jiexi: well, defaults to the defaultTimeout which defaults to 30000ms

Perhaps we can inline the 30000 here instead without the constructor param for now?

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
jiexi, line 1: did this file change besides whitespace?

packages/connect-evm/src/provider.ts
adonesky1: Also it seems like the point of adding this check in this layer at all is just for the validation that the chain is configured. I was actually thinking we want that validation for **all** methods not just readonly methods

packages/connect-evm/src/types.ts
```diff
@@ -27,6 +27,7 @@ export type MetamaskConnectEVMOptions = {
   core: MultichainCore;
   eventHandlers?: EventHandlers;
   notificationQueue?: unknown[];
+  readOnlyRpcMap?: Record<string, string>;
```
adonesky1: Lets type this more strictly 🙏 
```suggestion
  readOnlyRpcMap?: Record<CaipChainId, string>;
```

packages/connect-evm/src/provider.ts
jiexi: Yeah that makes sense to me. I know technically we can still handle requests to the wallet without a configured rpc endpoint for the chain, but let's simplify things by making the rule that you must provide a rpc endpoint for chain or you can't make _any_ requests to

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
```diff
@@ -1,95 +1,120 @@
...
-	 */
-	private async handleWithSdkState(options: InvokeMethodOptions): Promise<Json> {
-		// TODO: to be implemented
-		console.warn(`Method "${options.request.method}" is configured for SDK state handling, but this is not yet implemented. Falling back to wallet passthrough.`);
-		return this.handleWithWallet(options);
-	}
+  /**
+   * Responds directly from the SDK's session state.
+   */
+  private async handleWithSdkState(
```
adonesky1: No a part of your PR, but as I'm looking through this again I'm wondering is there actually a case for this on the multichain pacakge or is this really just something we use for managing selected chainId and selected account state on the EVM layer?

packages/connect-evm/src/connect.ts
adonesky1: hmmm 🤔 good question. I suppose that could work. Do you think we should accept both?

adonesky1, line 635: Can we rename this from `readonlyRpcMap` (everywhere in the connect-monorepo) to `supportedChains` so that it describes the fact that it is required to add all chains you want to use (and then the type describes the rpcUrl part)?

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
jiexi: Yeah makes sense. Let's leave that to the EVM layer. We already resolve permissions from sdk state at the Multichain layer

packages/connect-evm/src/connect.ts
jiexi: Sure, but not as part of this PR imo

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
ffmcgee725: good catch, reverted here
https://github.com/MetaMask/connect-monorepo/pull/33/commits/7d89af4dc94f68566ef001210d99ca3b35137d3b

packages/connect-multichain/src/multichain/rpc/handlers/rpcClient.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/33/commits/2ce835efd1e029b9bf0f5179f3d836545db31af8

packages/connect-evm/src/types.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/33/commits/e09e7f0937a1bc19f8977c6920d2624964950b90

packages/connect-evm/src/provider.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/33/commits/e1f2c2c3e01935823dfcf71c00a21afc7065e4d9

packages/connect-evm/src/connect.ts
ffmcgee725: IMO we should discourage accepting both, as it doesn't make for best API design principles. I do see the point of accepting Hex and then converting into CAIP-2, since we are handling EVM only in this layer.

It does make for more code to be maintained on our side though. Let me know your final thoughts on this and I can proceed accordingly.

ffmcgee725: PR updating this : https://github.com/MetaMask/connect-monorepo/pull/37

playground/legacy-evm-react-vite-playground/src/App.tsx
```diff
@@ -16,11 +17,25 @@ function useSDK() {
...
+      const infuraApiKey = import.meta.env.VITE_INFURA_API_KEY || '';
+      const readOnlyRpcMap = infuraApiKey
+        ? getInfuraRpcUrls(infuraApiKey)
+        : {
+            // Fallback public RPC endpoints if no Infura key is provided
+            'eip155:1': 'https://eth.llamarpc.com',
+            'eip155:5': 'https://goerli.infura.io/v3/demo',
+            'eip155:11155111': 'https://sepolia.infura.io/v3/demo',
+            'eip155:137': 'https://polygon-rpc.com',
+          };
```
jiexi, line 29: still feels weird to me to have to specify the chains in CAIP-2 format when this is an EVM only version of the SDK

jiexi (approved):

playground/legacy-evm-react-vite-playground/src/App.tsx
adonesky1: agreed. We can address this in a follow up

adonesky1 (approved):
LGTM

PR #42, opened 2025-11-18
refactor: simplify `MultichainOptions` configuration

ffmcgee725:
## Explanation

`MultichainOptions` currently includes flags `preferDesktop` and `preferExtension`, and we agreed as a team that naming convention for these is confusing as at first glance they seem to imply the same thing.

After analyzing `preferDesktop`, it was figured out that this flag actually is meant to indicate when the installation modal for the extension should be rendered instead of deeplinking user directly into mobile app (in case the extension is not installed in the browser)

That being said, this PR suggests renaming this config flag from `preferDesktop` to `showInstallModal` for less confusion.

We also bring back the change from this [previously reverted commit](https://github.com/MetaMask/connect-monorepo/pull/21/commits/14294ff3634290e766b982c450c27c0ba2b63500) to make sure the modal is properly rendered in Vite apps, as this was previously broken.

### **Before**

https://github.com/user-attachments/assets/711a2062-e5ad-4da4-9ad8-af860775159d

### **After**

https://github.com/user-attachments/assets/617bafc7-8742-4971-b38f-15d294658f3f

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-832

## Checklist

### DIFF
--- a/packages/connect-multichain/src/domain/ui/factory.ts
+++ b/packages/connect-multichain/src/domain/ui/factory.ts
@@ -23,7 +23,7 @@ export type ModalFactoryOptions = Pick<
   ui: {
     headless?: boolean; // Whether to run without UI
     preferExtension?: boolean; // Whether to prefer browser extension
-    showInstallModal?: boolean; // Whether to prefer desktop wallet
+    preferDesktop?: boolean; // Whether to render installation modal for desktop extension
   };
   onConnection: (
     transport: Transport,
--- a/packages/connect-multichain/src/domain/ui/factory.ts
+++ b/packages/connect-multichain/src/domain/ui/factory.ts
@@ -23,7 +23,7 @@ export type ModalFactoryOptions = Pick<
   ui: {
     headless?: boolean; // Whether to run without UI
     preferExtension?: boolean; // Whether to prefer browser extension
-    preferDesktop?: boolean; // Whether to render installation modal for desktop extension
+    showInstallModal?: boolean; // Whether to render installation modal for desktop extension
   };
   onConnection: (
     transport: Transport,
--- a/packages/connect-multichain/src/domain/ui/factory.ts
+++ b/packages/connect-multichain/src/domain/ui/factory.ts
@@ -23,7 +23,7 @@ export type ModalFactoryOptions = Pick<
   ui: {
     headless?: boolean; // Whether to run without UI
     preferExtension?: boolean; // Whether to prefer browser extension
-    preferDesktop?: boolean; // Whether to prefer desktop wallet
+    showInstallModal?: boolean; // Whether to prefer desktop wallet
   };
   onConnection: (
     transport: Transport,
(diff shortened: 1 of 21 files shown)


packages/connect-multichain/src/domain/ui/factory.ts
```diff
@@ -23,7 +23,7 @@ export type ModalFactoryOptions = Pick<
   ui: {
     headless?: boolean; // Whether to run without UI
     preferExtension?: boolean; // Whether to prefer browser extension
-    preferDesktop?: boolean; // Whether to prefer desktop wallet
+    preferDesktop?: boolean; // Whether to render installation modal for desktop extension
```
jiexi: should this be renamed too?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -558,11 +558,11 @@ export class MultichainSDK extends MultichainCore {
     await this.setupMWP();
 
     // Determine preferred option for install modal
-    const isDesktopPreferred = hasExtensionInstalled
-      ? preferDesktop
-      : !preferExtension || preferDesktop;
+    const shouldShowInstallModal = hasExtensionInstalled
+      ? showInstallModal
+      : !preferExtension || showInstallModal;
```
jiexi, line 563: Not feedback on your changes here, just thoughts in general.

So what the happens if extension isn't installed and you pass `showInstallModal: false` and `preferExtension: true`? No modal will be shown, but also there is nothing to connect to?

This ternary can be simplified to 
```
const shouldShowInstallModal = showInstallModal || (hasExtensionInstalled && !preferExtension);
```

but again `(hasExtensionInstalled && !preferExtension)` is weird. Really feels like this should be simplified all the way down to `const shouldShowInstallModal = showInstallModal`. I don't understand the intended behavior enough to say though

packages/connect-multichain/src/domain/ui/factory.ts
ffmcgee725: good catch ! https://github.com/MetaMask/connect-monorepo/pull/42/commits/51fa9f49c7dd6a8222e652edf0c150237d6f0947

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: > So what the happens if extension isn't installed and you pass `showInstallModal: false` and `preferExtension: true`? No modal will be shown, but also there is nothing to connect to?

You are asking the good questions, and as I've been saying lately as an answer to many of our queries in regards to this code, only god knows at this point 💀 

I do agree with everything you say in your ending statement, but at this point, not being familiar with the code paths in `connect-multichain` and how everything is wired, I do not feel confidence to fully break down and re-construct this logic to make more sense to us without the risk of breaking the existing behaviours. At least not until we make the release we are aiming for with stuff as stable as possible, and have proper e2e test coverage which will give us the confidence to refactor things properly.

jiexi (approved):

PR #46, opened 2025-11-25
feat: implement MM Connect Multichain Analytics

ffmcgee725:
## Explanation

This PR implements all of the remaining events/properties found in the [event’s schema](https://www.notion.so/metamask-consensys/MM-Connect-Analytics-Observability-2a9f86d67d68819283c4f9e2d4a62599?source=copy_link#266b49b5924f424bac236b2efc278bb6) that are not yet in the mm-connect package.

## Checklist


wenfix (dismissed):

packages/analytics/src/analytics.test.ts
```diff
@@ -65,19 +68,33 @@ t.describe('Analytics Integration', () => {
...
+      eventProperties.mmconnect_version,
+    );
+    analytics.setGlobalProperty('anon_id', eventProperties.anon_id);
+    analytics.setGlobalProperty('platform', eventProperties.platform);
+    analytics.setGlobalProperty(
+      'integration_type',
+      eventProperties.integration_type,
+    );
+    analytics.track('mmconnect_initialized', {
+      dapp_id: 'some-non-global-property',
```
adonesky1: Sorry if I missed this in our earlier call, but shouldn't this be a global property rather than a discreet event? Lets loop back w/ @adamceresko since I think we may want to fully remove the `mmconnect_initialized` event since I think it will be mostly noise (it will happen on every page load where MMConnect is installed)

ffmcgee725: After convo with @adamceresko, he preferred to keep this `initialized` event in, but you can sync and we loop back to this.

adonesky1: Ok. I'll loop back with him on this. But in any case I'd think `dapp_id` should be a global property no?

ffmcgee725: https://github.com/MetaMask/connect-monorepo/blob/66e8f8200277b018b7679f929eb0cf428b08ddf4/packages/connect-multichain/src/multichain/index.ts#L209C1-L210C1

☝🏾 it actually is, the test file property value is just misleading

ffmcgee725: changed here https://github.com/MetaMask/connect-monorepo/pull/46/commits/683fec5515757d490ae5ab52e08bbd3d82ec7956

wenfix (approved):

PR #57, opened 2025-12-02
Remove dangling branching logic where old sdk deeplink routes can be generated

adonesky1:
Per our sync this morning, we don't believe there are any cases where old deeplink format should be generated


packages/connect-evm/CHANGELOG.md
```diff
@@ -18,5 +18,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Typedoc generation ([#31](https://github.com/MetaMask/connect-monorepo/pull/31))
 - Permissioning accounts through `wallet_requestPermissions` ([#31](https://github.com/MetaMask/connect-monorepo/pull/31))
 - Enable conditional logging through options in `createMetamaskConnectEVM` ([#31](https://github.com/MetaMask/connect-monorepo/pull/31))
+- Fixed mobile deeplink bug when attempting to connect when already connected ([#57](https://github.com/MetaMask/connect-monorepo/pull/57))
```
adonesky1: we probably don't need fix changelog entries since this package has never been released 😅

jiexi (approved):

PR #60, opened 2025-12-04
fix: bump multichain api client and add event listener clean up for notifications

ffmcgee725:
## Explanation

In the Firefox Extension, when a dapp disconnects, the UI updates correctly, but attempting to reconnect does not prompt the wallet pop-up for permission granting.

This happens because [createDupeReqFilterStream middleware](https://github.com/MetaMask/metamask-extension/blob/4f6037d751e3a3b18d3b5019093dc3a9377ca223/app/scripts/lib/createDupeReqFilterStream.ts#L62) is responsible for keeping track of IDs that have been recently seen and not respond to request incoming from these matching IDs.

the proposed fix here is bumping `@metamask/multichain-api-client` to make sure we make use of the fix in [this PR](https://github.com/MetaMask/multichain-api-client/pull/91).

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-873

## Checklist

### DIFF
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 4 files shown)


wenfix (dismissed):
LGTM

wenfix (approved):

PR #63, opened 2025-12-04
feat: add connection id to simple deeplinks 

jiexi:
## Explanation

When deeplinking the user into the wallet for non-initial connection cases, the wallet has no idea which connection/session the user is opening the wallet from. In the case that the connection the dapp is using no longer exists on the wallet side, the user will sit in the wallet indefinitely waiting for a request that will never arrive.

This PR fixes that by including the connection id to the previously plain `metamask://connect/mwp` deeplink. It is now `metamask://connect/mwp?id=CONN_ID_HERE`. This will allow the wallet to prompt the user with an error toast if that connection no longer exists on the wallet side.

## References

See: https://github.com/MetaMask/metamask-mobile/pull/23700

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -486,9 +486,9 @@ export class MultichainSDK extends MultichainCore {
               },
             };
             const deeplink =
-              this.options.ui.factory.createDeeplink(connectionRequest);
+              this.options.ui.factory.createConnectionDeeplink(connectionRequest);
             const universalLink =
-              this.options.ui.factory.createUniversalLink(connectionRequest);
+              this.options.ui.factory.createConnectionUniversalLink(connectionRequest);
             if (this.options.mobile?.preferredOpenLink) {
               this.options.mobile.preferredOpenLink(deeplink, '_self');
             } else {
@@ -715,13 +715,19 @@ export class MultichainSDK extends MultichainCore {
     const shouldOpenDeeplink = secure && !showInstallModal;
 
     if (shouldOpenDeeplink) {
-      setTimeout(() => {
+      setTimeout(async () => {
+        const session = await this.transport.getActiveSession();
+        if (!session) {
+          throw new Error('No active session found');
+        }
+
+        const url = `${METAMASK_DEEPLINK_BASE}/mwp?id=${encodeURIComponent(session.id)}`;
         if (mobile?.preferredOpenLink) {
-          mobile.preferredOpenLink(METAMASK_DEEPLINK_BASE, '_self');
+          mobile.preferredOpenLink(url, '_self'); // here
         } else {
           openDeeplink(
             this.options,
-            METAMASK_DEEPLINK_BASE,
+            url,
             METAMASK_CONNECT_BASE_URL,
           );
         }
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -723,7 +723,7 @@ export class MultichainSDK extends MultichainCore {
 
         const url = `${METAMASK_DEEPLINK_BASE}/mwp?id=${encodeURIComponent(session.id)}`;
         if (mobile?.preferredOpenLink) {
-          mobile.preferredOpenLink(url, '_self'); // here
+          mobile.preferredOpenLink(url, '_self');
         } else {
           openDeeplink(
             this.options,
--- a/packages/connect-multichain/src/multichain/rpc/requestRouter.ts
+++ b/packages/connect-multichain/src/multichain/rpc/requestRouter.ts
@@ -61,7 +61,7 @@ export class RequestRouter {
 
 					const url = `${METAMASK_DEEPLINK_BASE}/mwp?id=${encodeURIComponent(session.id)}`;
 					if (mobile?.preferredOpenLink) {
-						mobile.preferredOpenLink(url, '_self'); // here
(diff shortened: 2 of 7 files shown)


packages/connect-multichain/src/multichain/rpc/requestRouter.ts
```diff
@@ -53,11 +53,17 @@ export class RequestRouter {
...
+				setTimeout(async () => {
+					const session = await this.transport.getActiveSession();
+					if (!session) {
+						throw new Error('No active session found');
+					}
+
+					const url = `${METAMASK_DEEPLINK_BASE}/mwp?id=${encodeURIComponent(session.id)}`;
 					if (mobile?.preferredOpenLink) {
-						mobile.preferredOpenLink(METAMASK_DEEPLINK_BASE, '_self');
+						mobile.preferredOpenLink(url, '_self'); // here
```
ffmcgee725:
```suggestion
						mobile.preferredOpenLink(url, '_self');
```

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -309,4 +310,8 @@ export class DefaultTransport implements ExtendedTransport {
       this.#notificationCallbacks.delete(callback);
     };
   }
+
+  getActiveSession(): Promise<Session | undefined> {
+    throw new Error('getActiveSession is purposely not implemented for the DefaultTransport');
```
ffmcgee725, line 318: should we leave a comment documenting the deeplink code path will never be triggered when using DefaultTransport, hence not needing an implementation for this method ?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -592,4 +585,16 @@ export class MWPTransport implements ExtendedTransport {
...
+
+  async getActiveSession(): Promise<Session | undefined> {
+    const { kvstore } = this;
+    const sessionStore = new SessionStore(kvstore);
+
+    try {
+      const [activeSession] = await sessionStore.list();
+      return activeSession;
+    } catch {
+      /* empty */
```
ffmcgee725: Since we are moving these parts at the moment, what do you think of adding a log here instead of leaving the catch empty ?
Can feel redundant since we are handling the `undefined` case upstream, but I would argue could also handle it here instead since this is reusable code that could be used on other parts of the codebase later on.

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -715,13 +715,19 @@ export class MultichainSDK extends MultichainCore {
...
+      setTimeout(async () => {
+        const session = await this.transport.getActiveSession();
+        if (!session) {
+          throw new Error('No active session found');
+        }
+
+        const url = `${METAMASK_DEEPLINK_BASE}/mwp?id=${encodeURIComponent(session.id)}`;
         if (mobile?.preferredOpenLink) {
-          mobile.preferredOpenLink(METAMASK_DEEPLINK_BASE, '_self');
+          mobile.preferredOpenLink(url, '_self'); // here
```
ffmcgee725:
```suggestion
          mobile.preferredOpenLink(url, '_self');
```

ffmcgee725 (commented):
Left some feedback, and also we should address the broken unit tests

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: Yes we should!

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: i'm not actually sure why this try catch is even needed, but it existed before

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: done here https://github.com/MetaMask/connect-monorepo/pull/63/commits/a9a72887a5306496bf8b7165d1c97ca9614b0cf7

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/63/commits/5854df45669ccdfd473e942529d0d68a55dd74a9

jiexi: left a log for now

jiexi:
no changelog because this is not developer or user facing

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -592,4 +585,18 @@ export class MWPTransport implements ExtendedTransport {
...
+
+  async getActiveSession(): Promise<Session | undefined> {
+    const { kvstore } = this;
+    const sessionStore = new SessionStore(kvstore);
+
+    try {
+      const [activeSession] = await sessionStore.list();
+      return activeSession;
+    } catch (error){
+      // TODO: verify if this try catch is necessary
```
adonesky1, line 597: Have we verified this?

adonesky1 (approved):

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: not yet. I looked at the code already, doesn't looks like this can throw, but not going to remove it out of abundance of safety

PR #75, opened 2025-12-09
fix: Fix initial deeplink connection rejection not updating connection status to disconnected

jiexi:
## Explanation

When making a deeplink connection, the rejection of the initial wallet_createSession request was not causing the dapp to correctly update it's local connection status to be "disconnected". This PR fixes that by updating the logic in the dappClient listener.

To test:
1. In the native browser
2. Connect
3. Accept the deeplink
4. Reject the approval
5. Return to the dapp
6. The dapp should show `status: disconnected` and allow you to attempt to reconnect again

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -454,7 +454,7 @@ export class MultichainSDK extends MultichainCore {
     return new Promise<void>(async (resolve, reject) => {
       // Handle the response to the initial wallet_createSession request
       const dappClientMessageHandler = (payload: unknown) => {
-        if (typeof payload !== 'object' || payload === null || !('data' in payload) {
+        if (typeof payload !== 'object' || payload === null || !('data' in payload)) {
           return;
         }
         const data = payload.data as { result?: SessionData , error?: unknown };
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -453,8 +453,11 @@ export class MultichainSDK extends MultichainCore {
   ) {
     return new Promise<void>(async (resolve, reject) => {
       // Handle the response to the initial wallet_createSession request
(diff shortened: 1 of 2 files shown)


packages/connect-multichain/src/multichain/index.ts
```diff
@@ -452,25 +452,23 @@ export class MultichainSDK extends MultichainCore {
     caipAccountIds: CaipAccountId[],
   ) {
     return new Promise<void>(async (resolve, reject) => {
-      this.dappClient.on('message', (payload: any) => {
+      // Handle the response to the initial wallet_createSession request
+      const dappClientMessageHandler = (payload: any) => {
```
ffmcgee725: is it possible to type `payload` here ?
```suggestion
      const dappClientMessageHandler = (payload: { data: { result: SessionData , error?: unknown }}) => {
```

So that we don't need typecasting below ?

jiexi: it's not necessarily a wallet_createSession response, but this is a good idea and i'll add it 🙏

ffmcgee725 (dismissed):

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -452,25 +452,26 @@ export class MultichainSDK extends MultichainCore {
     caipAccountIds: CaipAccountId[],
   ) {
     return new Promise<void>(async (resolve, reject) => {
-      this.dappClient.on('message', (payload: any) => {
-        const data = payload.data as Record<string, unknown>;
+      // Handle the response to the initial wallet_createSession request
+      const dappClientMessageHandler = (payload: unknown) => {
```
jiexi, line 456: payload has to be `unknown` here to match the `.on('message', handler)` type

ffmcgee725 (approved):

adonesky1 (approved):
YES

PR #76, opened 2025-12-10
refactor: fix multiple `MultichainSDK` typing and remove some eslint ignore comments from module

ffmcgee725:
## Explanation

This PR proposes some refactoring without business logic changes to `packages/connect-multichain/src/multichain/index.ts` `MultichainSDK` so that many `eslint-disable` comments can be removed from top level of the file.

Some need to be kept for now, since there are type mismatch issues (`void` and `Promise<void>` returns) between callbacks in our codebase that need to `return Promise<void>`, passed into functions from `multichain-api-client` that expect callbacks to `return void`.

A couple of unit test files also needed to be updated accordingly since `Vitest` is unable to spy on private variables with the `#` prefix.

## References

*Fixes https://consensyssoftware.atlassian.net/browse/WAPI-866

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -549,6 +549,7 @@ export class MultichainSDK extends MultichainCore {
             logger('Error tracking connection_established event', error);
           }
         }
+        return undefined; // explicitly return `undefined` to avoid eslintpromise/always-return
       })
       .catch(async (error) => {
         this.state = 'disconnected';
@@ -575,7 +576,7 @@ export class MultichainSDK extends MultichainCore {
             logger('Error tracking connection failed/rejected event', error);
           }
         }
-        return Promise.reject(error);
+        throw error instanceof Error ? error : new Error(String(error));
       });
   }
 
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -419,8 +419,8 @@ export class MultichainSDK extends MultichainCore {
             }
           },
         )
-        .catch(() => {
-          // Errors are handled by the callbacks
+        .catch((error) => {
+          reject(error instanceof Error ? error : new Error(String(error)));
         });
     });
   }
@@ -454,14 +454,14 @@ export class MultichainSDK extends MultichainCore {
     scopes: Scope[],
     caipAccountIds: CaipAccountId[],
   ): Promise<void> {
-    return new Promise<void>((resolve) => {
+    return new Promise<void>((resolve, reject) => {
       this.dappClient.on('message', (payload: any) => {
         const data = payload.data as Record<string, unknown>;
         if (typeof data === 'object' && data !== null) {
           if ('method' in data && data.method === 'wallet_createSession') {
             if (data.error) {
               this.state = 'loaded';
-              throw data.error as Error;
+              return reject(data.error as Error);
             }
             // TODO: is it .params or .result?
             const session = ((data as any).params ??
@@ -514,7 +514,7 @@ export class MultichainSDK extends MultichainCore {
         .then(resolve)
         .catch(async (error) => {
           await this.storage.removeTransport();
-          throw error;
+          reject(error instanceof Error ? error : new Error(String(error)));
         })
         .finally(() => {
           if (timeout) {
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -348,6 +348,83 @@ export class MultichainSDK extends MultichainCore {
     };
   }
 
+  async #renderInstallModalAsync(
+    desktopPreferred: boolean,
+    scopes: Scope[],
+    caipAccountIds: CaipAccountId[],
+  ): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      // Use Connection Modal
+      this.options.ui.factory
+        .renderInstallModal(
+          desktopPreferred,
+          async () => {
+            if (
+              this.dappClient.state === 'CONNECTED' ||
+              this.dappClient.state === 'CONNECTING'
+            ) {
+              await this.dappClient.disconnect();
+            }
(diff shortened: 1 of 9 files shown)


packages/connect-multichain/tests/types.ts
```diff
@@ -23,8 +20,8 @@ export type NativeStorageStub = {
 };
 
 export type MockedData = {
-  initSpy: t.MockInstance<MultichainSDK['init']>;
-  setupAnalyticsSpy: t.MockInstance<MultichainSDK['setupAnalytics']>;
+  initSpy: t.MockInstance<any>;
+  setupAnalyticsSpy: t.MockInstance<any>;
```
ffmcgee725, line 24: Vitest is unable to spy on private variables with the prefix `#`, so we need to be more loose here in order to test logic associated to them

jiexi: i don't think this is a vitest issue, but a js engine "issue". It shouldn't be possible to access `#` private methods or properties at all, even from the tests

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -359,106 +342,126 @@ export class MultichainSDK extends MultichainCore {
...
+              reject(error);
+            } else {
+              await this.storage.setTransport(TransportType.MWP);
+              resolve();
+            }
+          },
+        )
+        .catch(() => {
+          // Errors are handled by the callbacks
+        });
```
adonesky1: We still need to call reject on errors thrown in `renderInstallModal` before the callbacks are invoked 
```suggestion
      .catch((error) => {
          reject(
            error instanceof Error ? error : new Error(String(error)),
          );
        });
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -359,106 +342,126 @@ export class MultichainSDK extends MultichainCore {
...
+  ): Promise<void> {
+    return new Promise<void>((resolve) => {
       this.dappClient.on('message', (payload: any) => {
         const data = payload.data as Record<string, unknown>;
         if (typeof data === 'object' && data !== null) {
           if ('method' in data && data.method === 'wallet_createSession') {
             if (data.error) {
               this.state = 'loaded';
-              return reject(data.error);
+              throw data.error as Error;
```
adonesky1: Why did you remove this `reject` and throw instead? I think this will result in an uncaught exception since this event handler is not in the promise chain and so it won't be caught in #handleConnection

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -486,28 +493,28 @@ export class MultichainSDK extends MultichainCore {
...
-      this.transport
+      return this.transport
         .connect({ scopes, caipAccountIds })
         .then(resolve)
-        .catch((error) => {
-          this.storage.removeTransport();
-          reject(error);
+        .catch(async (error) => {
+          await this.storage.removeTransport();
+          throw error;
```
adonesky1: Same thing here... I think we still need to call `reject` so that the outer promise here doesn't hang?

adonesky1: Nested promises break my brain

jiexi: I think it still works the same if you throw, but let's keep the `reject`

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -359,106 +342,126 @@ export class MultichainSDK extends MultichainCore {
...
-              })
-              .catch((error) => {
-                if (error instanceof ProtocolError) {
-                  // Ignore Request expired errors to allow modal to regenerate expired qr codes
-                  if (error.code !== ErrorCode.REQUEST_EXPIRED) {
+                  });
+                },
+              );
+
+              (async (): Promise<void> => {
```
jiexi, line 385: is this necessary? can we make the function used to instantiate `new Promise` async?

i.e. `return new Promise<void>(async (resolve, reject) => {`

packages/connect-multichain/src/multichain/index.ts
jiexi: i think Alex is right here. If the errors are indeed handled by the callbacks, then this catch should never trigger in the first place. Either we should remove the catch and allow the errors (if any) to bubble up, or we should throw here

adonesky1 (commented):
Left a few comments about error handling but mostly looks great! Great work and thanks for proactively hopping on this

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: ☝🏾 that gives us back  `Promise executor functions should not be async.`, and removing the `async` keyword  makes it so calls to `this.transport.connect` need to be "callback hell'd" with the `then` and `catch` syntax which makes it harder to follow through in my opinion, so I prefer to just use this less hassling syntax so we can take advantage of `async/await` without Typescript screaming in our faces.

I do believe at one point we need to re-visit the rules established for this repo and adapt them to our preferences.

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: Typescript might have been giving me some slack which lead me to do the change, but can't remember now.
https://github.com/MetaMask/connect-monorepo/pull/76/commits/7e6ef36b900f6c1cabd6091aa1ec0789dcfd8bbe

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/76/commits/7e6ef36b900f6c1cabd6091aa1ec0789dcfd8bbe

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/76/commits/7e6ef36b900f6c1cabd6091aa1ec0789dcfd8bbe

adonesky1 (approved):
LGTM

PR #77, opened 2025-12-10
feat: support multichain initial connections

wenfix:
## Explanation

Allows requesting permissions for multiple chains during initial connection. Also enables prompting an `addEthereumChain` if the requested chain isn't configured in the wallet.

## Testing Steps

Requires [this commit on wagmi](https://github.com/wenfix/wagmi/pull/1/commits/c83189e92900c3451d6aa979cca6aec8be9dafa0)

### Case A
1. Go to wagmi's dapp
2. Select the chain you want to connect
3. Click metamask to connect -> all pre-configured chains in the wallet should appear in the permissions request
4. After accepting, all permissions should've been granted and the `chainId` in the dapp and wallet should match the requested chainId

### Case B
1. Go to wagmi's dapp
2. Select Celo (assuming its not configured in your wallet)
3. Click metamask to connect -> same as above, without Celo -> Accept it
4. There will be a follow-up prompt to add the Celo chain 
5. After accepting, all permissions should've been granted and the `chainId` in the dapp and wallet should be Celo's (`42220`).

## References
Closes [WAPI-880](https://consensyssoftware.atlassian.net/browse/WAPI-880)

> [!IMPORTANT]
> Had to do [this change](574ba677f700fc9d1a01fb6330df7c9c82c2fe1c) because importing this in wagmi was causing an issue. 

https://github.com/user-attachments/assets/f17e5cc5-db8e-460d-a390-3ccfbf7f8543

## Checklist

### DIFF
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -111,7 +111,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         const chainId = connectParams?.chainId ?? DEFAULT_CHAIN_ID;
         const withCapabilities = connectParams?.withCapabilities;
 
-        const availableChainIds = config.chains.map((chain) => chain.id);
+        const chainIds = config.chains.map((chain) => chain.id);
 
         const instance = await ensureMetamask();
 
@@ -131,15 +131,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
               if (parameters.connectAndSign) {
                 signResponse = await instance.connectAndSign({
                   message: parameters.connectAndSign,
-                  chainId,
-                  availableChainIds,
+                  chainIds,
                 });
               } else if (parameters.connectWith) {
                 connectWithResponse = await instance.connectWith({
                   method: parameters.connectWith.method,
                   params: parameters.connectWith.params,
-                  chainId,
-                  availableChainIds,
+                  chainIds,
                 });
               }
 
@@ -150,7 +148,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
               const result = await instance.connect({
                 chainId,
                 account: undefined,
-                availableChainIds,
+                chainIds,
               });
               accounts = result.accounts.map((account) => getAddress(account));
             }
--- a/integrations/wagmi/metamask-connector.ts
+++ b/integrations/wagmi/metamask-connector.ts
@@ -2,20 +2,20 @@ import {
   createMetamaskConnectEVM,
   type EIP1193Provider,
   type MetamaskConnectEVM,
-} from '@metamask/connect-evm'
+} from '@metamask/connect-evm';
 
 import {
   ChainNotConfiguredError,
   createConnector,
   ProviderNotFoundError,
-} from '@wagmi/core'
+} from '@wagmi/core';
 
 import type {
   Compute,
   ExactPartial,
   OneOf,
   Omit as StrictOmit,
-} from '@wagmi/core/internal'
+} from '@wagmi/core/internal';
 
 import {
   type Address,
@@ -28,38 +28,39 @@ import {
   type AddEthereumChainParameter as ViemAddEthereumChainParameter,
   withRetry,
   withTimeout,
-} from 'viem'
+} from 'viem';
 
 type CreateMetamaskConnectEVMParameters = Parameters<
   typeof createMetamaskConnectEVM
->[0]
+>[0];
 
-const DEFAULT_CHAIN_ID = 1
+const DEFAULT_CHAIN_ID = 1;
 
-export type MetaMaskParameters = {
-  dapp?: CreateMetamaskConnectEVMParameters['dapp'] | undefined
-} & OneOf<
-  | {
-      /* Shortcut to connect and sign a message */
-      connectAndSign?: string | undefined
-    }
-  | {
-      // TODO: Strongly type `method` and `params`
-      /* Allow `connectWith` any rpc method */
-      connectWith?: { method: string; params: unknown[] } | undefined
-    }
->
+export type MetaMaskParameters = Partial<
+  Pick<CreateMetamaskConnectEVMParameters, 'dapp' | 'debug'>
+> &
+  OneOf<
+    | {
+        /* Shortcut to connect and sign a message */
+        connectAndSign?: string | undefined;
+      }
+    | {
+        // TODO: Strongly type `method` and `params`
+        /* Allow `connectWith` any rpc method */
+        connectWith?: { method: string; params: unknown[] } | undefined;
+      }
+  >;
 
-metaMask.type = 'metaMask' as const
+metaMask.type = 'metaMask' as const;
 export function metaMask(parameters: MetaMaskParameters = {}) {
-  type Provider = EIP1193Provider
+  type Provider = EIP1193Provider;
   type Properties = {
-    onConnect(connectInfo: ProviderConnectInfo): void
-    onDisplayUri(uri: string): void
-  }
+    onConnect(connectInfo: ProviderConnectInfo): void;
+    onDisplayUri(uri: string): void;
+  };
 
-  let metamask: MetamaskConnectEVM | undefined
-  let metamaskPromise: Promise<MetamaskConnectEVM> | undefined
+  let metamask: MetamaskConnectEVM | undefined;
+  let metamaskPromise: Promise<MetamaskConnectEVM> | undefined;
 
   return createConnector<Provider, Properties>((config) => {
     // Helper to ensure metamask is initialized
@@ -71,7 +72,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
               `eip155:${chain.id}`,
               chain.rpcUrls.default?.http[0],
             ]),
-          )
+          );
 
           metamaskPromise = createMetamaskConnectEVM({
             dapp: parameters.dapp ?? { name: window.location.hostname },
@@ -84,11 +85,12 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
             api: {
               supportedNetworks,
             },
-          })
+            debug: parameters.debug,
+          });
         }
-        metamask = await metamaskPromise
+        metamask = await metamaskPromise;
       }
-      return metamask
+      return metamask;
     }
 
     const connector = {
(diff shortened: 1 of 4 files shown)


packages/connect-evm/src/connect.ts
```diff
@@ -261,26 +276,32 @@ export class MetamaskConnectEVM {
...
     logger('request: connect', { chainId, account });
-    const caipChainId: Scope[] = chainId ? [`eip155:${chainId}`] : [];
 
-    const caipAccountId: CaipAccountId[] =
-      chainId && account ? [`eip155:${chainId}:${account}`] : [];
+    const caipChainIds = Array.from(
+      new Set(
+        availableChainIds?.concat(DEFAULT_CHAIN_ID) ?? [DEFAULT_CHAIN_ID],
+      ),
+    ).map((id) => `eip155:${id}`);
```
ffmcgee725, line 292: I understand having the logic as such while we are the only consumers of our lib and are responsible for implementing the wagmi connector changes.

But if a consumer passes in a `chainId`, and `availableChainIds`, we should perhaps enforce that `availableChainIds` includes `chainId` (and throw an error otherwise)
And also change this logic so that the defaulted array is `[chainId]` instead of `[DEFAULT_CHAIN_ID]`

packages/connect-evm/src/connect.ts
```diff
@@ -261,26 +276,32 @@ export class MetamaskConnectEVM {
...
 
-    await this.#core.connect(caipChainId, caipAccountId, forceRequest);
+    const caipAccountIds =
+      chainId && account
+        ? caipChainIds.map((caipChainId) => `${caipChainId}:${account}`)
+        : [];
+
+    await this.#core.connect(
+      caipChainIds as Scope[],
+      caipAccountIds as CaipAccountId[],
```
ffmcgee725, line 300: If we type the variables above, we have no need for this type cast

wenfix: Not really, both `map` functions will return `string[]` so you either have to cast them there or here, and here felt easier to read.

packages/connect-evm/src/connect.ts
wenfix: Good catch!

wenfix: For the last part tho, we always want to have mainnet (`DEFAULT_CHAIN_ID`) because we can't be certain `chainId` exists in the wallet. This way, we can always connect, regardless of the parameter `chainId`.

adonesky1: > For the last part tho, we always want to have mainnet (DEFAULT_CHAIN_ID) because we can't be certain chainId exists in the wallet. This way, we can always connect, regardless of the parameter chainId.

@ffmcgee725 yeah as @wenfix says here we want to always include ethereum mainnet in the request to ensure that a connection can be established even if none of the requested chainIds are currently configured in the wallet, that way the MWP transport will succeed and the dapp can make an addChain request for the chain it needs. Because of our architectural coupling of the transport session and the permission sessions we have to make sure the connection succeeds this way

integrations/wagmi/metamask-connector.ts
```diff
@@ -98,82 +100,102 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
+        chainId?: number | undefined;
+        isReconnecting?: boolean | undefined;
+        withCapabilities?: withCapabilities | boolean | undefined;
       }) {
-        const chainId = connectParams?.chainId ?? DEFAULT_CHAIN_ID
-        const withCapabilities = connectParams?.withCapabilities
+        const chainId = connectParams?.chainId ?? DEFAULT_CHAIN_ID;
+        const withCapabilities = connectParams?.withCapabilities;
+
+        const availableChainIds = config.chains.map((chain) => chain.id);
```
adonesky1: I think a more appropriate name for this variable is `dappSupportedNetworks` or `dappSupportedChainIds` or `dappConfiguredChainIds`?

packages/connect-evm/src/connect.ts
```diff
@@ -261,26 +276,40 @@ export class MetamaskConnectEVM {
    * @param options.chainId - The chain ID to connect to (defaults to 1 for mainnet)
```
adonesky1:
```suggestion
   * @param options.chainId - The chain ID to initially select upon connection (defaults to 1 for mainnet)
```

packages/connect-evm/src/connect.ts
```diff
@@ -261,26 +276,40 @@ export class MetamaskConnectEVM {
...
-    {
-      chainId,
-      account,
-      forceRequest,
-    }: {
-      chainId: number;
-      account?: string | undefined;
-      forceRequest?: boolean;
-    } = { chainId: DEFAULT_CHAIN_ID }, // Default to mainnet if no chain ID is provided
+    { chainId, account, forceRequest, availableChainIds }: ConnectOptions = {
```
adonesky1:
```suggestion
    { initiallySelectedChainId, account, forceRequest, dappSupportedChainIds }: ConnectOptions = {
```
Open to other names, but I think these need to be more specific and explain themselves better

packages/connect-evm/src/connect.ts
```diff
@@ -295,23 +324,28 @@ export class MetamaskConnectEVM {
       accounts: initialAccounts.result as Address[],
     });
 
-    this.#core.transport.onNotification((notification) => {
-      // @ts-expect-error TODO: address this
-      if (notification?.method === 'metamask_accountsChanged') {
+    // Remove previous notification handler if it exists
+    this.#removeNotificationHandler?.();
```
adonesky1, line 318: can you explain why this change was necessary?

adonesky1 (commented):
Tested and looks good!

Just left a few variable name changes and one question!

packages/connect-evm/src/connect.ts
```diff
@@ -261,26 +276,40 @@ export class MetamaskConnectEVM {
...
+      Array.isArray(availableChainIds) &&
+      !availableChainIds.includes(chainId)
+    ) {
+      throw new Error('availableChainIds must include the provided chainId');
+    }
 
-    await this.#core.connect(caipChainId, caipAccountId, forceRequest);
+    const caipChainIds = Array.from(
+      new Set(
+        availableChainIds?.concat(DEFAULT_CHAIN_ID) ?? [DEFAULT_CHAIN_ID],
```
jiexi: `[DEFAULT_CHAIN_ID, ...availableChainIds]`?

jiexi: spreading undefined is safe

packages/connect-evm/src/connect.ts
jiexi: it's because he's overwriting `this.#removeNotificationHandler` below, so he calls it (a function that unregisters the listener from the transport) before doing so

jiexi: ahh, sorry. previously we were not cleaning up this listener here if the case that `connect` is called multiple times. @wenfix has done so here

packages/connect-evm/src/connect.ts
jiexi: i feel like we shouldn't expose an availableChainIds and should instead just include Mainnet as the other chain to connect to in addition to the chain specified by `chainId`.

The goal here is just to make sure a connection is establish-able, so it's not strictly required to request all of the chains the dapp supports at this time.

I'm thinking about this more from the API standpoint of not wanting to expose both `chainId` AND `availableChainIds` though. I am open to the alternative of making `chainId` plural `chainIds`, and same for `account` to `accounts` while we're at it. This will match the interface for `MultichainSDK.connect` more closely

jiexi: I imagine this makes it so this is less of a drop-in replacement for SDKv1, but IMO that isn't important as the developer adopting this will need to make other changes anyway and this is not a major change to have to adopt

packages/multichain-ui/package.json
```diff
@@ -37,6 +37,16 @@
...
+    "./dist/loader/index.js": {
+      "import": {
+        "types": "./dist/loader/index.d.ts",
+        "default": "./dist/loader/index.js"
+      },
+      "require": {
+        "types": "./dist/loader/index.d.ts",
+        "default": "./dist/loader/index.cjs.js"
+      }
+    },
```
jiexi: reminder to try to remove this again and see exactly where the failure is occuring

packages/connect-evm/src/connect.ts
adonesky1: > The goal here is just to make sure a connection is establish-able, so it's not strictly required to request all of the chains the dapp supports at this time.

I think you're conflating two goals. One of the goals is definitely to request all the chains the dapp supports up front rather than requiring that they add permission each time the user switches to a new unpermissioned chain. This flow is particularly painful in mobile. The other goal is - as you say - to make sure the connection can be established no matter which chains are requested. We're attempting to solve both of these here.

I agree that the interface is not ideal. I think the compromise here may be to make just one `chainIds` array, and the first chainId in the array is the one we take to be the initially selected chainId.

adonesky1: @jiexi [agrees with ](https://consensys.slack.com/archives/C08SWDEN2GG/p1765391475112489?thread_ts=1765387622.461849&cid=C08SWDEN2GG):

> I think the compromise here may be to make just one chainIds array, and the first chainId in the array is the one we take to be the initially selected chainId.

packages/multichain-ui/package.json
jiexi: Screenshot 2025-12-10 at 3 13 00 PM

was able to see QR code / install modal working with this removed

packages/connect-evm/src/connect.ts
wenfix: I disagree with this approach, I think to a consumer of this package that will be more confusing as we'd be mixing 2 different things into a weird abstraction that'll force consumers to read the docs to understand whats going on. Its not intuitive to me and I think having them separate gives us more clarity. 

I agree its not ideal and its not very ergonomic as you have to pass the same `chainId` in 2 parameters but I think its more clear. Open to continue the discussion ofc.

packages/connect-evm/src/connect.ts
wenfix: Only on an object. On an array it requires the value to be `iterable` otherwise it throws a `TypeError`.

packages/connect-evm/src/connect.ts
wenfix: 673c9fc2

packages/connect-evm/src/connect.ts
wenfix: var name fix here: 673c9fc2

integrations/wagmi/metamask-connector.ts
wenfix: 673c9fc2

packages/multichain-ui/package.json
wenfix: done here: a316d9df

packages/connect-evm/src/connect.ts
ffmcgee725: but notice how prior we did not typecast, we merely type the variable explicitly, and it goes through successfully
```ts
   const caipChainId: Scope[] = chainId ? [`eip155:${chainId}`] : [];
   const caipAccountId: CaipAccountId[] =
      chainId && account ? [`eip155:${chainId}:${account}`] : [];
```

Are you 100% sure we cannot maintain it as such ?

packages/connect-evm/src/connect.ts
jiexi: oh neat

packages/connect-evm/src/connect.ts
```diff
@@ -258,29 +273,48 @@ export class MetamaskConnectEVM {
...
       account,
       forceRequest,
-    }: {
-      chainId: number;
-      account?: string | undefined;
-      forceRequest?: boolean;
-    } = { chainId: DEFAULT_CHAIN_ID }, // Default to mainnet if no chain ID is provided
+      dappSupportedChainIds,
+    }: ConnectOptions = {
+      initiallySelectedChainId: DEFAULT_CHAIN_ID,
```
adonesky1: Don't we need dappSupportedChainIds to also have a default value in this case?

adonesky1 (approved):

jiexi (approved):

PR #79, opened 2025-12-10
fix: addEthereumChain not being called on mobile after switchChain fails

wenfix:
## Checklist

### DIFF
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -461,8 +461,9 @@ export class MetamaskConnectEVM {
 
       // When using the MWP transport, the error is returned instead of thrown,
       // so we force it into the catch block here.
-      if (result?.error) {
-        throw new Error(result.error.message);
+      const resultWithError = result as { error?: { message: string } };
+      if (resultWithError?.error) {
+        throw resultWithError.error;
       }
 
       await this.#trackWalletActionSucceeded(method, scope, params);
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -463,12 +463,12 @@ export class MetamaskConnectEVM {
       // so we force it into the catch block here.
       const resultWithError = result as { error?: { message: string } };
       if (resultWithError?.error) {
-        throw resultWithError.error;
+        throw new Error(resultWithError.error.message);
       }
 
       await this.#trackWalletActionSucceeded(method, scope, params);
 
-      if ((result as unknown as { result: unknown }).result === null) {
+      if ((result as { result: unknown }).result === null) {
         // result is successful we eagerly call onChainChanged to update the provider's selected chain ID.
         this.#onChainChanged(hexChainId);
       }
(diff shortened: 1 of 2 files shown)


packages/connect-evm/src/connect.ts
```diff
@@ -458,10 +458,18 @@ export class MetamaskConnectEVM {
         method: 'wallet_switchEthereumChain',
         params,
       });
+
+      // When using the MWP transport, the error is returned instead of thrown,
+      // so we force it into the catch block here.
+      if (result?.error) {
+        throw new Error(result.error.message);
```
jiexi: is it possible just to rethrow the error as is without instantiating a new error from the previous errors message?

adonesky1: done here: https://github.com/MetaMask/connect-monorepo/pull/79/commits/7f9960507dbec217d8b938d926ed1e1f222470c9

adonesky1:
Sound on 🔉 
https://github.com/user-attachments/assets/fd168479-d816-4049-b186-f621bea258a1

oof my microphone doesn't sound great, but I think you get the idea.

1. Wondering if we can get the deeplink prompt to go away when you return to the wallet
2. Other deeplinks seem to be broken after the add flow... not sure if its causal. But I'm not able to reproduce the issue without first going through the addChain flow so possibly causal.

jiexi:
`#addEthereumChain` probably needs to fire an chainChanged event (and update it's internal selected chainId) after `wallet_addEthereumChain` returns successfully (like we do for wallet_switchEthereumChain)

wenfix:
> `#addEthereumChain` probably needs to fire an chainChanged event (and update it's internal selected chainId) after `wallet_addEthereumChain` returns successfully (like we do for wallet_switchEthereumChain)

I think this happens implicitly: 
1) addEthereumChain will switch the wallet's selected network into the added network
2) we're listening to `chainChanged` event and triggering our internal one, which will do what you described.

Am I missing something?

EDIT: after chatting w/ Alex I realized you probably meant eagerly so I'll update it!

wenfix:
> Sound on 🔉 https://github.com/user-attachments/assets/fd168479-d816-4049-b186-f621bea258a1
> 
> oof my microphone doesn't sound great, but I think you get the idea.
> 
> 1. Wondering if we can get the deeplink prompt to go away when you return to the wallet
> 2. Other deeplinks seem to be broken after the add flow... not sure if its causal. But I'm not able to reproduce the issue without first going through the addChain flow so possibly causal.



https://github.com/user-attachments/assets/36e5f35b-acf9-40ad-9320-2c0174ada60b


I'm not seeing either of the issues you pointed out when using a physical android device. Could this be either iOS specific or simulator related? 

I don't understand how this change could've caused it as they're specifically on the `switchChain` method... 🤔 

(There seems to be some wonkyness as a error toast shows up during the confirmation 🤔 )

wenfix:
had to re-wrap the error due to the [only-throw-error eslint rule](https://typescript-eslint.io/rules/only-throw-error/) that's defined in [our eslint-configs](https://github.com/MetaMask/eslint-config/blob/0b8cf6db0edcc5e1582bf188a5495bfd6f1cf069/packages/typescript/src/index.mjs#L18).

jiexi (approved):

adonesky1 (approved):

PR #81, opened 2025-12-11
fix: clean up dangling event listeners after auditing event listeners

ffmcgee725:
## Explanation

This PR proposes a clean up to a couple dangling event listeners we had across `connect-multichain`, `connect-evm` and mwp transport.

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-885

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -511,6 +511,7 @@ export class MultichainSDK extends MultichainCore {
           reject(error);
         })
         .finally(() => {
+          this.dappClient.off('message', dappClientMessageHandler);
           if (timeout) {
             clearTimeout(timeout);
           }
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -508,10 +508,10 @@ export class MultichainSDK extends MultichainCore {
         .then(resolve)
         .catch((error) => {
           this.storage.removeTransport();
+          this.dappClient.off('message', dappClientMessageHandler);
           reject(error);
         })
         .finally(() => {
-          this.dappClient.off('message', dappClientMessageHandler);
           if (timeout) {
             clearTimeout(timeout);
           }
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.ts
@@ -428,11 +428,12 @@ export class MWPTransport implements ExtendedTransport {
     });
 
     return connectionPromise
-      .catch(() => {
+      .catch((error) => {
         if (initialConnectionMessageHandler) {
           this.dappClient.off('message', initialConnectionMessageHandler);
           initialConnectionMessageHandler = undefined;
         }
+        throw error;
       })
       .finally(() => {
         if (timeout) {
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.ts
@@ -374,13 +374,13 @@ export class MWPTransport implements ExtendedTransport {
                     messagePayload.method === 'wallet_createSession' ||
                     messagePayload.method === 'wallet_sessionChanged'
                   ) {
-                    if (initialConnectionMessageHandler) {
-                      this.dappClient.off(
-                        'message',
-                        initialConnectionMessageHandler,
-                      );
-                    }
                     if (messagePayload.error) {
+                      if (initialConnectionMessageHandler) {
+                        this.dappClient.off(
+                          'message',
+                          initialConnectionMessageHandler,
+                        );
+                      }
                       return rejectConnection(messagePayload.error);
                     }
                     await this.storeWalletSession(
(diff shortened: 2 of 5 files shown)


packages/connect-multichain/src/multichain/index.ts
```diff
@@ -511,6 +511,7 @@ export class MultichainSDK extends MultichainCore {
           reject(error);
         })
         .finally(() => {
+          this.dappClient.off('message', dappClientMessageHandler);
```
jiexi: we might only want this in the catch. If it's in the finally, we might unsubscribe before the event listener has processed a wallet_createSession response?.. @adonesky1

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -369,6 +374,12 @@ export class MWPTransport implements ExtendedTransport {
                     messagePayload.method === 'wallet_createSession' ||
                     messagePayload.method === 'wallet_sessionChanged'
                   ) {
+                    if (initialConnectionMessageHandler) {
+                      this.dappClient.off(
+                        'message',
+                        initialConnectionMessageHandler,
+                      );
+                    }
```
jiexi: I think when @adonesky1 and I looked at this, we may have discovered that this listener was the only one also listening for wallet_sessionChanged, caching the value and/or notifying callbacks for it.

We need to verify subsequent wallet_sessionChanged events still come through

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -407,6 +431,10 @@ export class MWPTransport implements ExtendedTransport {
       if (timeout) {
         clearTimeout(timeout);
       }
+      if (initialConnectionMessageHandler) {
+        this.dappClient.off('message', initialConnectionMessageHandler);
+        initialConnectionMessageHandler = undefined;
+      }
```
jiexi: i don't think we want to do this in the finally since the wallet_sessionChanged may not have arrived at this point

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: i don't think `handleMessage` is written correctly to be a stand in here

ffmcgee725: I can remove this unsubscription from here then if wallet_sessionChanged depends on it, but we run into an issue where this handler never removed, meaning if `connect()` is called multiple times, multiple handlers accumulate (each call adds a new one).

also seems like `handleMessage` is processing the same `wallet_sessionChanged` notifications, but that method is such a confusion pile I'm not even sure what to say regarding this.

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/81/commits/5217c8298272880b393967b4f4da192476cb6f71 moved into catch clause

packages/connect-multichain/src/multichain/transports/mwp/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/81/commits/5217c8298272880b393967b4f4da192476cb6f71 moved removal of listener if payload has error, but might be open to issue I stated above

packages/connect-multichain/src/multichain/transports/mwp/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/81/commits/5217c8298272880b393967b4f4da192476cb6f71 moved into catch clause

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -391,23 +404,42 @@ export class MWPTransport implements ExtendedTransport {
...
+                rejectConnection(error);
+              });
           },
         );
       }
 
       timeout = setTimeout(() => {
+        if (initialConnectionMessageHandler) {
+          this.dappClient.off('message', initialConnectionMessageHandler);
+        }
```
jiexi: will the reject below cause this to end up in the catch on line 431? As in do we need this here or can we rely on that one?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -327,6 +327,9 @@ export class MWPTransport implements ExtendedTransport {
     }
 
     let timeout: NodeJS.Timeout;
+    let initialConnectionMessageHandler:
+      | ((message: unknown) => Promise<void>)
+      | undefined;
     const connectionPromise = new Promise<void>((resolve, reject) => {
```
jiexi, line 333: oof okay these nested promises made understanding actual logic difficult.

`connection` doesn't need to exist. `resumeResolve` and `resumeReject` and just be `resolve` and `reject`. You don't need to do this as part of this PR though if you don't want since it's not what you're trying to change.

You are actually right in this comment

https://github.com/MetaMask/connect-monorepo/pull/81/changes/BASE..926d2b71ae3ad98ca58928bc0841648629e4c695#r2611792396

You should only need this logic in the finally of the outer most `connectionPromise.finally()` as you did before.
```
                      if (initialConnectionMessageHandler) {
                        this.dappClient.off(
                          'message',
                          initialConnectionMessageHandler,
                        );
                      }
```

The rest of these should not be needed unless you see something I'm missing

Sorry about that!

ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/81/commits/7f7c572ba96a7c628fdb112cb86846da90497ea4

packages/connect-multichain/src/multichain/transports/mwp/index.ts
ffmcgee725: Good catch, addressed here https://github.com/MetaMask/connect-monorepo/pull/81/commits/7f7c572ba96a7c628fdb112cb86846da90497ea4 , but moved from `catch` clause into `finally` which will also execute in case `reject` is called.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: thank you!

jiexi (approved):

adonesky1 (approved):
LGTM

PR #84, opened 2025-12-12
improve mwp connection management

adonesky1:
### Problem

The `MWPTransport.connect()` method could leave the underlying `DappClient` stuck in a `CONNECTING` state, causing subsequent connection attempts to fail with `"Cannot connect when state is CONNECTING"`.

### Root Causes

1. **No guard against concurrent connection calls** – Multiple calls to `connect()` would attempt to call `dappClient.connect()` simultaneously
2. **Message handler leak** – The `'message'` handler registered during connection was never removed, causing memory leaks and duplicate handlers on reconnection
3. **Unhandled resume failures** – If `dappClient.resume()` threw, the `'connected'` event listener would hang indefinitely
4. **Missing await on `onResumeSuccess`** – Errors from this async function were not properly caught

### Changes

- Add `connectionPromise` to track in-progress connections and return the existing promise on concurrent calls
- Track and clean up the connection message handler (`connectionMessageHandler`) on success, failure, timeout, and disconnect
- Handle `resume()` errors by removing the `'connected'` listener and rejecting the promise
- Properly await `onResumeSuccess()` with `.catch()` to surface errors

### DIFF
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.ts
@@ -85,6 +85,16 @@ export class MWPTransport implements ExtendedTransport {
 
   private windowFocusHandler: (() => void) | undefined;
 
+  /**
+   * Tracks the current connection promise to prevent concurrent connection attempts
+   */
+  private connectionPromise: Promise<void> | undefined;
+
+  /**
+   * Cleanup function for the message handler registered during fresh connections
+   */
+  private connectionMessageHandler: ((message: unknown) => void) | undefined;
+
   get pendingRequests() {
     return this.__pendingRequests;
   }
@@ -319,6 +329,12 @@ export class MWPTransport implements ExtendedTransport {
     scopes: Scope[];
     caipAccountIds: CaipAccountId[];
   }): Promise<void> {
+    // Guard against concurrent connection attempts - return existing promise if one is in progress
+    if (this.connectionPromise) {
+      logger('connection already in progress, waiting for existing promise');
+      return this.connectionPromise;
+    }
+
     const { dappClient } = this;
 
     const session = await this.getActiveSession();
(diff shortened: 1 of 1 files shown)


packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -369,6 +405,15 @@ export class MWPTransport implements ExtendedTransport {
...
                   ) {
+                    // Clean up this handler once we get the response
+                    if (this.connectionMessageHandler) {
+                      this.dappClient.off(
+                        'message',
+                        this.connectionMessageHandler,
+                      );
+                      this.connectionMessageHandler = undefined;
+                    }
+
```
jiexi, line 416: can we move all cleanup to the very last finally on line 468?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -391,22 +438,39 @@ export class MWPTransport implements ExtendedTransport {
                   data: request,
                 },
               })
-              .catch(rejectConnection);
+              .catch((err: Error) => {
+                // Clean up handler on connection failure
+                if (this.connectionMessageHandler) {
```
jiexi, line 443: same about moving this into the very last finally

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -319,6 +329,12 @@ export class MWPTransport implements ExtendedTransport {
     scopes: Scope[];
     caipAccountIds: CaipAccountId[];
   }): Promise<void> {
+    // Guard against concurrent connection attempts - return existing promise if one is in progress
+    if (this.connectionPromise) {
+      logger('connection already in progress, waiting for existing promise');
+      return this.connectionPromise;
+    }
+
```
jiexi, line 337: The connection promise will eventually timeout, so this won't get stuck forever, but I thought one of our goals was to avoid having to wait for that to occur before being allowed to attempt to establish a new connection

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: i think this also needs to go into the `disconnect()`

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: i suppose we have delegated this to the `disconnect()` and `connect()` in the layer above this

PR #85, opened 2025-12-12
fix: Fix stuck connection by attempting to clear any old state before new connection attempt

jiexi:
## Explanation

Fixes stuck MWP connections that cannot be resolved using the disconnect and connect buttons in the wagmi react demo. This is partly because wagmi does not actually call `disconnect()` on the MM Connect instance if it has not self reported to be connected. This leaves the user soft locked if their MWP connection fails to establish as pressing connect with try to reuse the existing broken connection.

This PR fixes that by calling `disconnect()` internally if the connection status is anything but `connected` thus theoretically getting into a clean state before establishing a whole new connection.

## Checklist


adonesky1 (approved):
🙏 🙏 🙏

PR #89, opened 2026-01-05
fix: Exclude multichain-ui from React Native builds

adonesky1:
**Problem:**
[Commit `cd90fc0`](https://github.com/MetaMask/connect-monorepo/commit/cd90fc01c7c4dceb6ab94cb8b395ed1d27cfd0bf#diff-7b8d6ce63337a0017c757570ce5d506299e7eb3e318173df3b0ad0c599317eeeL26-R34) introduced a static top-level import of `@metamask/multichain-ui/loader`, which causes Metro (React Native's bundler) to fail. Metro cannot process Stencil's dynamically generated imports with template literal variables.

**Solution:**
Move the Stencil import into platform-specific preload files and swap them at build time:

- `preload.web.ts` - Loads Stencil custom elements (browser builds)
- `preload.native.ts` - No-op function (React Native builds)
- `tsup.config.ts` - esbuild plugin swaps `preload.web` → `preload.native` for RN builds

**Result:**
React Native bundle no longer contains any `@metamask/multichain-ui` references. Browser builds continue to work unchanged.


playground/multichain-react-native-playground/package.json
jiexi, line 1: was this meant to be included in this PR?

adonesky1: nope whoops

packages/connect-multichain/tsup.config.ts
```diff
@@ -147,6 +148,17 @@ export default defineConfig([
...
+      // Swap web preload (with Stencil) for native preload (no-op) in React Native builds
+      {
+        name: 'replace-preload-web',
+        setup(build) {
+          build.onResolve({ filter: /preload\.web$/ }, (args) => {
+            return { path: path.resolve(args.resolveDir, './preload.native.ts') };
+          });
+        },
+      },
+    ],
```
jiexi, line 161: rather than doing it this way, what are your thoughts on specifiying index entrypoints via package.json?

`packages/connect-multichain/src/ui/index.ts` -> `packages/connect-multichain/src/ui/index.(index|native|browser).ts` ?

I know this would require more code reorg though

playground/multichain-react-native-playground/package.json
adonesky1: https://github.com/MetaMask/connect-monorepo/pull/89/commits/f4b74a0fa000af5c2e9aa1aa7dc6faf8986fc236

packages/connect-multichain/tsup.config.ts
adonesky1: Yeah I considered this but thought this was the least invasive change. Happy to make this change though since it would be cleaner

jiexi (approved):

PR #93, opened 2026-01-06
fix: Fix `MultichainSDK.create()` getting blocked by connection resumption

jiexi:
## Explanation

Fixes an issue with `MultichainSDK.create()` not returning until any pre-existing connection have been re-established. This is an issue for MWP (specifically on android simulator on chrome, we're not sure why) where the `#setupTransport()` method called via `#init()` from `create()` takes a long time before failing. We don't see the same problem on physical android devices, on ios simulator safari, nor on ios physical device chrome. It seems exclusive to android simulator.

Another potential way to tackle this bug would be to figure out why the MwpTransport.connect() takes so long to resolve on Android simulator but not other environments, but that still leaves the question why we are blocking the `create()` method on reconnection. 

**Repro steps:**
1. On Android Simulator Chrome, open the wagmi integration test dapp
2. Connect, open the deeplink, but do not accept or reject it in the wallet
3. Return to the dapp
4. Refresh the page
5. Afterwards, disconnect/connect buttons will seem unresponsive until the initial connection resumption attempt resolves

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -264,7 +264,6 @@ export class MultichainSDK extends MultichainCore {
         logger('MetaMaskSDK: init already initialized');
       } else {
         await this.#setupAnalytics();
-        await this.#setupTransport();
         if (this.options.analytics?.enabled) {
           try {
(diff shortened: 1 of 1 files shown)


jiexi:
This change causes resume from refresh to be broken

PR #102, opened 2026-01-08
Add legacy compatibility methods and ensure stable method binding for EIP1193Provider

adonesky1:
This PR enhances `EIP1193Provider` with legacy JSON-RPC compatibility methods and ensures all public methods maintain their `this` binding when extracted or passed as callbacks.

**Changes**

1. **Legacy compatibility methods** (`types.ts`, `provider.ts`)
   - Added `chainId` getter (alias for `selectedChainId`)
   - Added `sendAsync()` for callback/promise-based JSON-RPC requests
   - Added `send()` for callback-based JSON-RPC requests
   - Added corresponding `JsonRpcRequest`, `JsonRpcResponse`, and `JsonRpcCallback` types

2. **Method binding in constructor** (`provider.ts`)
   - Bound all public methods (`request`, `sendAsync`, `send`) and inherited EventEmitter methods (`on`, `off`, `emit`, `once`, `removeListener`, `listenerCount`) to `this` in the constructor

**Why this is necessary**

Many dApps and wallet integration libraries pass provider instances through adapters, store them in state, or destructure methods:

```javascript
const { request } = provider;
request({ method: 'eth_accounts' }); // Without binding, `this` is undefined
```

Because `EIP1193Provider` uses private fields (`#core`, `#accounts`), losing the `this` binding causes runtime errors like `Cannot read from private field`. Binding methods in the constructor makes the provider "extraction-safe" and eliminates the need for consumers to wrap the provider in a Proxy.

### DIFF
--- a/packages/connect-evm/src/provider.ts
+++ b/packages/connect-evm/src/provider.ts
@@ -8,6 +8,9 @@ import type {
   Address,
   EIP1193ProviderEvents,
   Hex,
+  JsonRpcCallback,
+  JsonRpcRequest,
+  JsonRpcResponse,
   ProviderRequest,
   ProviderRequestInterceptor,
 } from './types';
@@ -32,6 +35,21 @@ export class EIP1193Provider extends EventEmitter<EIP1193ProviderEvents> {
     super();
     this.#core = core;
     this.#requestInterceptor = interceptor;
+
+    // Bind all public methods to ensure `this` context is preserved
+    // when methods are extracted or passed as callbacks.
+    // This eliminates the need for Proxy wrappers in consumers.
+    this.request = this.request.bind(this);
+    this.sendAsync = this.sendAsync.bind(this);
+    this.send = this.send.bind(this);
+
+    // Bind inherited EventEmitter methods
+    this.on = this.on.bind(this);
+    this.off = this.off.bind(this);
+    this.emit = this.emit.bind(this);
+    this.once = this.once.bind(this);
+    this.removeListener = this.removeListener.bind(this);
+    this.listenerCount = this.listenerCount.bind(this);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/connect-evm/src/provider.ts
```diff
@@ -32,6 +35,21 @@ export class EIP1193Provider extends EventEmitter<EIP1193ProviderEvents> {
...
+    this.sendAsync = this.sendAsync.bind(this);
+    this.send = this.send.bind(this);
+
+    // Bind inherited EventEmitter methods
+    this.on = this.on.bind(this);
+    this.off = this.off.bind(this);
+    this.emit = this.emit.bind(this);
+    this.once = this.once.bind(this);
+    this.removeListener = this.removeListener.bind(this);
+    this.listenerCount = this.listenerCount.bind(this);
```
jiexi, line 52: thoughts on just switching to the fat arrow syntax for declaring the methods then?

Example
```
 sendAsync = <TParams = unknown, TResult = unknown> (
    request: JsonRpcRequest<TParams>,
    callback?: JsonRpcCallback<TResult>,
  ): Promise<JsonRpcResponse<TResult> | void> => {
...
```

jiexi: i think the above pattern breaks inheritance though. Just want to avoid us adding a new method later and forgetting to also bind it

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -277,6 +277,11 @@ export class MultichainSDK extends MultichainCore {
       this.state = 'pending';
       logger('MetaMaskSDK error during initialization', error);
     }
+
+    this.#setupTransport().catch(async () => {
+      await this.storage.removeTransport();
+      this.state = 'pending';
+    });
```
jiexi: need to undo this?

adonesky1: whoops yep! https://github.com/MetaMask/connect-monorepo/pull/102/commits/80ec06f82b84946c9aa424436997077ff75f146d

packages/connect-evm/src/provider.ts
adonesky1: Lets stick with this more verbose binding strategy for now, I don't think we'll need to add more methods but if we do it should be pretty clear given the comments what's wrong if you forget to bind

jiexi (approved):

PR #103, opened 2026-01-08
fix: Fix dangling sessions when initial connection is attempted but the wallet doesn't respond timely or at all

jiexi:
## Explanation

Fixes an issue where the sdk hangs indefinitely in the connecting state when:
1. The user initiates a connection to the wallet but neither accepts/rejects it
2. The user initiates a connection to the wallet and accepts/rejects it, but the response isn't received by the sdk in a timely manner and hangs (but is recoverable if the user refreshes the dapp)

## Checklist


packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -646,4 +631,34 @@ export class MWPTransport implements ExtendedTransport {
       return undefined;
     }
   }
+
+  private async waitForWalletSessionIfNotCached() {
```
adonesky1, line 663: Maybe worth adding a brief inline comment explaining why this is necessary?

jiexi: https://github.com/MetaMask/connect-monorepo/pull/103/commits/c1e0066a9a8a63f2f282a162c8bf060e91b2003c

adonesky1 (approved):
LGTM!

PR #108, opened 2026-01-14
fix: incorrect chainId on page refresh

wenfix:
## Explanation

When a user switches to a chain other than Ethereum Mainnet and refreshes the page, the selected chainId incorrectly reverts to chainId 1. This occurred because the code assumed the first permitted chain in the session scopes was the selected chain, rather than querying the wallet for the actual selected chain.

## Solution

Query the wallet for the currently selected chain using `eth_chainId` instead of assuming the first permitted chain is selected. This ensures the correct chainId persists across page refreshes.

## Changes

- Added private method `#getSelectedChainId()` that queries `eth_chainId` from the wallet and validates it against permitted chains
- Updated `connect()` and `#attemptSessionRecovery()` to use the new method
- Falls back to the first permitted chain if `eth_chainId` fails or returns an invalid chain

## References

Fixes [WAPI-924](https://consensyssoftware.atlassian.net/browse/WAPI-924)

## After

https://github.com/user-attachments/assets/4647a551-95a6-43eb-b703-f07121b784a0

## Checklist

### DIFF
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -256,6 +256,32 @@ export class MetamaskConnectEVM {
   }
 
   /**
+   * Gets the currently selected chainId from the wallet, or falls back to the first permitted chain.
+   *
+   * @param permittedChainIds - Array of permitted chain IDs in hex format
+   * @returns The selected chainId (hex string)
+   */
+  async #getSelectedChainId(permittedChainIds: Hex[]): Promise<Hex> {
+    // Query the wallet for the currently selected chainId instead of assuming the first permitted chain
+    try {
+      const chainIdResponse = await this.#core.transport.sendEip1193Message<
+        { method: 'eth_chainId'; params: [] },
+        { result: string; id: number; jsonrpc: '2.0' }
+      >({ method: 'eth_chainId', params: [] });
+      const walletChainId = chainIdResponse.result as Hex;
+      // Validate that the returned chainId is in the permitted chains list
+      if (permittedChainIds.includes(walletChainId)) {
+        return walletChainId;
+      }
+    } catch (error) {
+      logger('Error querying eth_chainId', error);
+    }
+
+    // Fallback to the first permitted chain if eth_chainId failed or returned an invalid chain
+    return permittedChainIds[0];
+  }
+
+  /**
    * Connects to the wallet with the specified chain ID and optional account.
    *
    * @param options - The connection options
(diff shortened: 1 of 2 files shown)


packages/connect-evm/src/connect.ts
```diff
@@ -255,6 +255,32 @@ export class MetamaskConnectEVM {
...
+   * @returns The selected chainId (hex string)
+   */
+  async #getSelectedChainId(permittedChainIds: Hex[]): Promise<Hex> {
+    // Query the wallet for the currently selected chainId instead of assuming the first permitted chain
+    try {
+      const chainIdResponse = await this.#core.transport.sendEip1193Message<
+        { method: 'eth_chainId'; params: [] },
+        { result: string; id: number; jsonrpc: '2.0' }
+      >({ method: 'eth_chainId', params: [] });
+      const walletChainId = chainIdResponse.result as Hex;
```
ffmcgee725:
```suggestion
      const chainIdResponse = await this.#core.transport.sendEip1193Message<
        { method: 'eth_chainId'; params: [] },
        { result: Hex; id: number; jsonrpc: '2.0' }
      >({ method: 'eth_chainId', params: [] });
      const walletChainId = chainIdResponse.result;
```

Can we do this ? that way, no typecast necessary

packages/connect-evm/src/connect.ts
```diff
@@ -771,14 +798,16 @@ export class MetamaskConnectEVM {
...
+
+      const chainId = await this.#getSelectedChainId(permittedChainIds);
 
       if (permittedChainIds.length && permittedAccounts.result) {
         this.#onConnect({
-          chainId: permittedChainIds[0],
+          chainId,
           accounts: permittedAccounts.result as Address[],
         });
       }
```
ffmcgee725, line 813:
```suggestion
      const permittedAccounts = await this.#core.transport.sendEip1193Message<
        { method: 'eth_accounts'; params: [] },
        { result: Address[]; id: number; jsonrpc: '2.0' }
      >({ method: 'eth_accounts', params: [] });

      const chainId = await this.#getSelectedChainId(permittedChainIds);

      if (permittedChainIds.length && permittedAccounts.result) {
        this.#onConnect({
          chainId,
          accounts: permittedAccounts.result,
        });
      }
```

same as above

ffmcgee725 (dismissed):

ffmcgee725 (approved):

PR #113, opened 2026-01-15
fix: cache last selected chainId

wenfix:
## Explanation

When a user switches to a chain other than Ethereum Mainnet and refreshes the page, the selected chainId incorrectly reverts to chainId 1. This occurred because the code assumed the first permitted chain in the session scopes was the selected chain, rather than persisting the user's actual selection.

### Solution

Cache the selected chainId in storage whenever it changes, and retrieve it from cache during session recovery and connection. This ensures the correct chainId persists across page refreshes.

## Checklist

### DIFF
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -41,6 +41,7 @@ import {
 } from './utils/type-guards';
 
 const DEFAULT_CHAIN_ID = 1;
+const CHAIN_STORE_KEY = 'cache_eth_chainId';
 
 /** The options for the connect method */
 type ConnectOptions = {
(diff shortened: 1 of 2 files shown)


packages/connect-evm/src/connect.ts
```diff
@@ -256,28 +257,36 @@ export class MetamaskConnectEVM {
...
+        // The cached value might be stored as a number (from metamask_chainChanged) or hex string (from eth_chainId)
+        let chainId: Hex;
+        try {
+          const parsed = JSON.parse(cachedChainId);
+          chainId =
+            typeof parsed === 'number' ? numberToHex(parsed) : (parsed as Hex);
+        } catch {
+          // If parsing fails, try treating it as a direct hex string
+          chainId = cachedChainId as Hex;
+        }
```
ffmcgee725: Do we need all this fail-safe logic to retrieve the cached value ? 

Looking below at where we `adapter.set` the value, we are always storing an `Hex` value, so all we need to do here is just fetch it as is, or am I missing something ?

wenfix: 74eda93d

ffmcgee725 (approved):

PR #115, opened 2026-01-15
feat: implement wagmi connector support in browser playground

ffmcgee725:
## Explanation

This PR updates the playground to use Wagmi via the local connector file only (not via NPM package).

https://github.com/user-attachments/assets/13723d3f-9de6-4577-8bce-2d6bfb7286db

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-918

## Checklist


chakra-guy (approved):
LGTM, but I think it would be nice to add some docs around why we need this and how this works. Maybe also a cursor rule that can catch itself when we work with this.

PR #121, opened 2026-01-20
fix: Resolve Buffer polyfill requirement for browser/React Native consumers

chakra-guy:
## Explanation

**The Problem:**
Consumers integrating `@metamask/connect-evm` or `@metamask/connect-multichain` in browser environments (Vite, CRA, React Native) encounter a runtime error: `Cannot read properties of undefined (reading 'isBuffer')`. The root cause is `eciesjs` (used for MWP encryption), which internally calls `Buffer.from()`. 

This created an **undocumented, implicit requirement**.

**The Solution:**
This PR transforms that implicit requirement into an **explicit, automatically-handled dependency**. The package now ships with the Buffer polyfill built-in, meaning consumers get a working integration out-of-the-box with zero configuration.

**How it works:**
- A lightweight shim (`src/polyfills/buffer-shim.ts`) sets up `globalThis.Buffer` before any dependent code runs
- The `buffer` package is added as a dependency but marked external, allowing npm/yarn to deduplicate it with other packages that may already include it
- Browser and React Native entry points import the shim first

**What this means for consumers:**
- No more manual bundler configuration required
- No more cryptic "isBuffer" errors
- Works out-of-the-box in Vite, CRA, and React Native
- If they already have Buffer polyfilled, the dependency is shared (no duplication)

## References

- https://consensyssoftware.atlassian.net/browse/WAPI-921

## Checklist

### DIFF
--- a/integrations/wagmi/src/main.tsx
+++ b/integrations/wagmi/src/main.tsx
@@ -2,15 +2,12 @@ import { createSyncStoragePersister } from '@tanstack/query-sync-storage-persist
 import { QueryClient } from '@tanstack/react-query'
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
 import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client'
-import { Buffer } from 'buffer'
 import React from 'react'
 import ReactDOM from 'react-dom/client'
 import { deserialize, serialize, WagmiProvider } from 'wagmi'
 
 import './index.css'
 
-// `@coinbase-wallet/sdk` uses `Buffer`
-globalThis.Buffer = Buffer
 
 import App from './App.tsx'
 import { config } from './wagmi.ts'
--- a/packages/connect-multichain/src/index.browser.ts
+++ b/packages/connect-multichain/src/index.browser.ts
@@ -1,3 +1,6 @@
+// Buffer polyfill must be imported first to set up globalThis.Buffer
+import './polyfills/buffer-shim';
+
 import type { CreateMultichainFN, StoreClient } from './domain';
 import { MultichainSDK } from './multichain';
 import { Store } from './store';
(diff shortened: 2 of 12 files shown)


ffmcgee725 (dismissed):

integrations/wagmi/src/main.tsx
jiexi, line 1: should the wagmi package.json also remove `buffer`?

packages/connect-multichain/src/index.browser.ts
```diff
@@ -1,3 +1,6 @@
+// Buffer polyfill must be imported first to set up globalThis.Buffer
+import './polyfills/buffer-shim';
+
```
jiexi, line 3: do we need to verify the built js files don't have this reordered incorrectly?

jiexi:
> If they already have Buffer polyfilled, the dependency is shared (no duplication)

The dependency would potentially be duplicated in the install node_modules though i think

jiexi:
Thoughts on making Buffer a peerDep and devDep rather than a regular Dependency?

integrations/wagmi/src/main.tsx
chakra-guy: Good catch, yes that should be removed.

packages/connect-multichain/src/index.browser.ts
chakra-guy: Verified! I checked the built output and the order is correct. esbuild preserves import order from entry points, so since index.browser.ts imports the shim first, it executes before any code that depends on Buffer.

chakra-guy:
> > If they already have Buffer polyfilled, the dependency is shared (no duplication)
> 
> The dependency would potentially be duplicated in the install node_modules though i think

You're right that yarn will still install the buffer package. However, yarn does deduplicate at the node_modules level too.

chakra-guy:
> Thoughts on making Buffer a peerDep and devDep rather than a regular Dependency?

Yeah, but I feel like the main goal of this PR was to eliminate the implicit requirement that consumers had to figure out on their own. Making it a peerDep would kinda undo that, no?

ffmcgee725 (approved):

PR #123, opened 2026-01-20
feat: solana support

wenfix:
## Explanation

Introduces `@metamask/connect-solana`, a new package that wraps `@metamask/solana-wallet-standard` with MetaMask Connect, enabling Solana dapps to use MetaMask via the wallet-standard protocol. It also adds playground support for solana-wallet-standard which enables testing this feature.

### Testing Steps

1. Install dependencies and build (`yarn install && yarn build`)
2. Launch browser playground (`yarn workspace @metamask/browser-playground start`)
3. Click `Connect (Solana)` button (ensure no extension is installed if using desktop browser)
4. **IMPORTANT: Make sure `Solana Devnet` is permitted when accepting**
5. Confirm connection -> Solana Card should appear in dapp
6. Try interactions available interactions such as sign, which should succeed.

## References

Fixes [WAPI-751](https://consensyssoftware.atlassian.net/browse/WAPI-754)

## Checklist

[WAPI-751]: https://consensyssoftware.atlassian.net/browse/WAPI-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wenfix:
@SocketSecurity ignore @solana/wallet-standard-core
@SocketSecurity ignore @solana/wallet-standard-features
@SocketSecurity ignore @solana/wallet-standard-util

`solana-devs` is a known publisher of Solana related libraries.

wenfix:
@SocketSecurity ignore npm/@solana/wallet-standard-chains@1.1.1
@SocketSecurity ignore npm/@solana/wallet-standard-core@1.1.2
@SocketSecurity ignore npm/@solana/wallet-standard-features@1.3.0
@SocketSecurity ignore npm/@solana/wallet-standard-util@1.1.2


`solana-devs` is a known publisher of Solana related libraries.

packages/connect-solana/src/connect.ts
```diff
@@ -0,0 +1,60 @@
...
+ * const wallet = client.getWallet();
+ * ```
+ */
+export async function createSolanaClient(
+  options: SolanaConnectOptions,
+): Promise<SolanaClient> {
+  const core = await createMultichainClient({
+    dapp: options.dapp,
+    api: {
+      supportedNetworks: options.api?.supportedNetworks ?? {},
```
adonesky1: Maybe we add mainnet as the default?
```suggestion
      supportedNetworks: options.api?.supportedNetworks ?? {
      'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp': 'https://api.mainnet-beta.solana.com'
      },
```

packages/connect-solana/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@metamask/connect-solana",
+  "version": "0.1.0",
```
jiexi: this needs to start at `0.0.0`

packages/connect-solana/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Registers the MetaMask wallet with the wallet-standard registry.
+   * This makes MetaMask automatically discoverable by Solana dapps.
+   * @param walletName - Optional custom name for the wallet
+   */
+  registerWallet: (walletName?: string) => Promise<void>;
+  /**
+   * Disconnects from the wallet and revokes the session.
+   */
+  disconnect: () => Promise<void>;
+};
```
jiexi, line 66: the EVM and Solana Clients are pretty different at this point. I wonder if we should try to establish a common pattern. For example should we remove the `request()` method from ConnectEvm and make the 1193 provider the only way to make requests like we are kind of doing here? (i know MultichainCore is exposed though)

packages/connect-solana/CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- Initial release ([#123](https://github.com/MetaMask/connect-monorepo/pull/123))
```
jiexi: technically we aren't releasing it yet. sorry to nit pick lol

packages/connect-solana/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+   * @param walletName - Optional custom name for the wallet
+   * @returns The wallet instance
+   */
+  getWallet: (walletName?: string) => Wallet;
+  /**
+   * Registers the MetaMask wallet with the wallet-standard registry.
+   * This makes MetaMask automatically discoverable by Solana dapps.
+   * @param walletName - Optional custom name for the wallet
+   */
+  registerWallet: (walletName?: string) => Promise<void>;
```
jiexi, line 61: we should align on if MMC or the dapp is responsible for determining if an announcement event should be made here and for EVM

packages/connect-solana/src/types.ts
```diff
@@ -0,0 +1,48 @@
+import type { MultichainCore } from '@metamask/connect-multichain';
+import type { Wallet } from '@wallet-standard/base';
+
+/**
+ * Configuration options for creating a Solana client.
+ */
+export type SolanaConnectOptions = {
```
jiexi: should this be derived more directly from ConnectMultichain types instead like we do for the EVM client?

```
Pick<MultichainOptions, 'dapp' | 'api'> &
  { ui?: Omit<MultichainOptions['ui'], 'factory'>; } & {
    eventHandlers?: Partial<EventHandlers>;
    debug?: boolean;
  }
```

packages/connect-solana/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+    /** The name of the dapp */
+    name: string;
+    /** The URL of the dapp */
+    url?: string;
+    /** The icon URL of the dapp */
+    iconUrl?: string;
+  };
+  /** Optional API configuration */
+  api?: {
+    /** A map of CAIP chain IDs to RPC URLs for supported networks */
```
jiexi: should this be `Solana Genesis Hash to RPC URLs`?

packages/connect-solana/src/connect.ts
jiexi: IMO we should not expect the `solana` namespace to be included in the callers `options.api.supportedNetworks` key values

jiexi: ^ in reference to how there is no massaging going on here. Not about Donesky's comment.

playground/browser-playground/craco.config.js
jiexi, line 1: why are the changes here needed? Sorry if obvious 🙏

package.json
```diff
@@ -88,6 +88,10 @@
     "typescript-eslint": "^8.10.0",
     "yargs": "^17.7.2"
   },
+  "resolutions": {
+    "@metamask/utils": "^11.8.1",
```
jiexi: why did this get moved to the root level package.json?

and why the bs58 forced resolution?

packages/connect-solana/src/connect.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    });
+  });
+
+  it('should return core instance from createMultichainClient', async () => {
+    const client = await createSolanaClient(mockOptions);
+
+    expect(client.core).toBe(mockCore);
+  });
+
+  it('should get wallet using getWalletStandard', async () => {
```
jiexi: thoughts on moving the tests starting from here and down into their own scenarios? everything above this is definitely more related to `createSolanaClient` but i don't think the ones starting here and below are

playground/browser-playground/src/components/SolanaWalletCard.tsx
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    setLoading(true);
+    setError(null);
+    setSignedMessage(null);
+
+    try {
+      const encodedMessage = new TextEncoder().encode(message);
+      const signature = await signMessage(encodedMessage);
+      setSignedMessage(Buffer.from(signature).toString('base64'));
```
jiexi, line 37: doesn't really matter but is there a way to do this without Buffer?

jiexi: again, doesn't really matter though

playground/browser-playground/src/App.tsx
```diff
@@ -119,6 +122,23 @@ function App() {
     }
   }, [customScopes, connectors, wagmiConnectAsync]);
 
+  const connectSolana = useCallback(async () => {
+    // Find the MetaMask wallet in registered wallets
+    const metamaskWallet = wallets.find((w) => 
+      w.adapter.name.toLowerCase().includes('metamask')
```
jiexi: is there a chance this will perfer the extension announced event instead of the ConnectSolana one?

packages/connect-solana/src/connect.ts
adonesky1: I think you might need to be a bit more clear about your followup comments here Jiexi I don't think I fully understand

packages/connect-solana/src/types.ts
adonesky1: Think we're going to need to sync on this one as well @jiexi, not sure I follow

packages/connect-solana/src/types.ts
adonesky1: 🤔 also not following this one haha

package.json
wenfix: both were (unsafe) attempts at reducing bundle size. I've removed them. 9e771ff4

packages/connect-solana/CHANGELOG.md
wenfix: e3bcd7a0

packages/connect-solana/package.json
wenfix: 0f0590de

playground/browser-playground/craco.config.js
wenfix: 1) Swapped `'require'` before `'import'` to prefer CommonJS over ESM. This is likely because some Solana packages (like `@solana/web3.js` or `@metamask/solana-wallet-standard`) have dual CJS/ESM exports, and their ESM builds may have compatibility issues with the Create React App / webpack setup. Preferring CommonJS avoids those issues.

2) This disables webpack's "fully specified" requirement for `.mjs` files from `node_modules`. Without this, webpack would require imports like `import foo from './bar.js'` (with extension), but many ESM packages use `import foo from './bar'` (without extension). Solana ecosystem packages often trigger this issue.

packages/connect-solana/src/connect.ts
wenfix: What I understand is that'd be preferable to just receive the RPC URL, perhaps mapped to one of `mainnet`, `devnet`, `testnet` instead of the Solana namespace?
Something like:  

```ts
{ 'mainnet': 'https://api.mainnet-beta.solana.com' }
``` 

And we do the namespace mapping ourselves. WDYT?

packages/connect-solana/src/types.ts
wenfix: e0f37b61

packages/connect-solana/src/connect.test.ts
wenfix: ea2d9d592fc9dd4e4de51d51e84b53bdb2cae0c4

packages/connect-solana/src/connect.ts
wenfix: Here's what I'm suggesting: ec248c81

packages/connect-solana/src/types.ts
jiexi: Right now it is the dapp's responsibility to call registerWallet() in order to make the MMC created provider discoverable rather than MMC automatically announcing itself. Does this seem like the right pattern? I was more on the of side of MMC automatically announcing itself being better, but kind of changing my mind now.

tldr; nevermind this is good!

packages/connect-solana/src/connect.ts
jiexi: ahh sorry for the ambiguity.

I mean that this logic might look more like
```
// no `solana:` prefix
options.api.supportedNetworks = {
  '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp': 'https://api.mainnet-beta.solana.com'
}

const supportedNetworksWithSolanaNamespace = Object.fromEntries(
  Object.entries(originalObject).map(([key, value]) => [
    `solana:${key}`,
    value
  ])
)

...

{
      supportedNetworks: supportedNetworksWithSolanaNamespace 
}
```

jiexi: i.e. supportedNetworks is expected just to be the genesis hash (which identifies the network/chain) without the `solana:` prefix which we will add ourselves before passing it to `createMultichainClient`

adonesky1:
Running into a build issue on the browser playground:

Screenshot 2026-01-28 at 2 27 09 PM

packages/connect-solana/src/connect.ts
adonesky1: yeah I suppose this makes sense though we currently for `connect-evm` we use the full caip chainIds -> https://github.com/MetaMask/connect-monorepo/blob/ec248c818d3c11fd9286cb8cd77f757ff244eb4f/packages/connect-evm/src/connect.ts#L941
&
https://github.com/MetaMask/connect-monorepo/blob/ec248c818d3c11fd9286cb8cd77f757ff244eb4f/packages/connect-multichain/src/domain/multichain/api/types.ts#L81-L85

playground/browser-playground/craco.config.js
adonesky1: Tried unswapping these to [resolve this issue](https://github.com/MetaMask/connect-monorepo/pull/123#issuecomment-3814301510) but it didn't help 😢

adonesky1: From Claude re [this issue](https://github.com/MetaMask/connect-monorepo/pull/123#issuecomment-3814301510)

### Three versions of bs58 → base-x:

| bs58 version | base-x version | Pulled in by |
|--------------|----------------|--------------|
| **4.0.1** | 3.0.11 (CJS only) | `@solana/web3.js`, `borsh` |
| **5.0.0** | 4.0.1 | `@solana-mobile/*` packages |
| **6.0.0** | 5.0.1 (ESM + CJS) | `@metamask/solana-wallet-standard`, `@reown/appkit`, `@walletconnect/utils` |

### Why this causes the error:

1. **Webpack hoists the oldest version** - `base-x@3.0.11` gets installed at the root `node_modules/base-x/` because it was likely resolved first or has the most dependents.

2. **Nested versions exist but webpack ignores them** - `base-x@5.0.1` is nested inside `node_modules/bs58/node_modules/base-x/`, but when webpack sees `import basex from 'base-x'`, it resolves from the root, not the nested location.

3. **ESM/CJS mismatch** - `bs58@6.0.0` (used by your Solana wallet adapter chain) has ESM code that does:
   ```js
   import basex from 'base-x';
   export default basex(ALPHABET);  // expects basex to be a function
   ```
   But `base-x@3.0.11` (the hoisted version) is CJS-only with a different export structure, so `basex` isn't a function.

wenfix:
> Running into a build issue on the browser playground:
> 
> Screenshot 2026-01-28 at 2 27 09 PM

Adding Solana introduced `bs58@6.0.0` (ESM) alongside `bs58@4.0.1` (CJS), causing webpack bundling failures due to ESM/CJS interoperability issues with the `base-x` dependency.

### Solution: 
Force all packages to use `bs58@4.0.1` via yarn resolution.


This is safe because the `bs58` API is identical across versions:
```ts
interface BS58 {  
  encode(source: Buffer | Uint8Array): string;  
  decode(string: string): Buffer;  // v4 returns Buffer  
  decodeUnsafe(string: string): Buffer | undefined;
}
```

The only difference is `decode()` return type:
v4/v5: Returns `Buffer`
v6: Returns `Uint8Array`

Since `Buffer` extends `Uint8Array`, code expecting `Uint8Array` works transparently with `Buffer`. All consumer packages (`@metamask/solana-wallet-standard`, `@solana/web3.js`, `@walletconnect/utils`, etc.) are compatible.

packages/connect-solana/src/connect.ts
jiexi: API be funkier, but these were always just meant to be convience wrappers around the underlying multichain client which we would rather people adopt and use directly instead. Let's keep as caip chain id

adonesky1:
I'm not seeing the `sign **and** send` flow working:

Each time I click it (even after repeated attempts) I get prompted to connect again (I think this has to do with devnet permissions not getting granted initially...? though I don't know why the connection request happens again and again then) but as you can see I don't ever get the subsequent signature request to show:

https://github.com/user-attachments/assets/c38f13d1-0ecd-4c85-8567-a7a45edf732d

wenfix:
> I'm not seeing the `sign **and** send` flow working:
> 
> Each time I click it (even after repeated attempts) I get prompted to connect again (I think this has to do with devnet permissions not getting granted initially...? though I don't know why the connection request happens again and again then) but as you can see I don't ever get the subsequent signature request to show:
> 
>  Screen.Recording.2026-01-29.at.2.08.38.PM.mov

Yes, I've confirmed its due to the connection request not including the Solana Devnet scope. If you manually include it, you should see the tx confirmation pop-up. Let's discuss some of this during our standup today.

adonesky1 (dismissed):
LGTM!

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -240,14 +243,6 @@ export class DefaultTransport implements ExtendedTransport {
...
       }
     } else if (!walletSession || options?.forceRequest) {
-      const optionalScopes = addValidAccounts(
-        getOptionalScopes(options?.scopes ?? []),
-        getValidAccounts(options?.caipAccountIds ?? []),
-      );
-      const createSessionParams: CreateSessionParams<RPCAPI> = {
-        optionalScopes,
-        sessionProperties: options?.sessionProperties,
-      };
```
adonesky1, line 250: Previously we weren't passing along sessionProperties in certain cases -> theoretically this could be its own PR but was causing an issue where Solana connection wasn't leading to additional network permissions being added as is expected (and currently happens in the injected script version)

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -206,6 +206,15 @@ export class DefaultTransport implements ExtendedTransport {
       throw new Error(sessionRequest.error.message);
     }
     let walletSession = sessionRequest.result as SessionData;
+
+    const createSessionParams: CreateSessionParams<RPCAPI> = {
+      optionalScopes: addValidAccounts(
+        getOptionalScopes(options?.scopes ?? []),
+        getValidAccounts(options?.caipAccountIds ?? []),
+      ),
+      sessionProperties: options?.sessionProperties,
```
ffmcgee725:
```suggestion
      sessionProperties: options?.sessionProperties ?? {},
```
if `options?.sessionProperties` is `undefined`, I believe we are required to pass in an empty object for `sessionProperties`

playground/browser-playground/src/App.tsx
```diff
@@ -227,16 +282,32 @@ function App() {
...
+              </button>
+            )}
+
+            {!solanaConnected && (
+              <button
+                type="button"
+                onClick={connectSolana}
+                className="bg-purple-500 text-white px-5 py-2 rounded text-base hover:bg-purple-600 transition-colors"
+              >
+                Connect (Solana)
```
ffmcgee725, line 300: this button needs a `data-testid` for e2e testability

eslint.config.mjs
```diff
@@ -78,7 +78,9 @@ const config = createConfig([
             'packages/*/*.config.ts',
             'playground/*/*.config.ts',
             'integrations/wagmi/wagmi.config.ts',
+            'packages/connect-solana/src/connect.test.ts',
           ],
+          maximumDefaultProjectFileMatchCount_THIS_WILL_SLOW_DOWN_LINTING: 12,
```
ffmcgee725, line 82: Why was this config necessary ? And why 12 ?

wenfix: This is a bit of a drag. `allowDefaultProject` is for linting files that aren't configured in `tsconfig`. Regardless of adding the test, since we added more config files, we've hit the limit, which is 8. Cursor said it was safe to raise this limit.

wenfix: Screenshot 2026-02-04 at 16 27 23

wenfix: Think this hints at something misconfigured but also think it falls out of the scope of this PR.

packages/connect-multichain/src/multichain/transports/default/index.ts
wenfix: This is typed with [CreateSessionParams<T extends RpcApi> from multichain-api-client](https://github.com/MetaMask/multichain-api-client/blob/decd68f641188720c81a7000dbcbcf1ae6bcbda5/src/types/multichainApi.ts#L83). 
Do we need to fix this type?

ffmcgee725: Nevermind this... Looking at [core handler](https://github.com/MetaMask/core/blob/281cb401fe12bb9949cc13c5aa0397acef34b13a/packages/multichain-api-middleware/src/handlers/wallet-createSession.ts#L100), seems like it's optional, and we only consider it invalid if no properties are actually passed in (but can come as `undefined` and then we default it into an empty object)

My memory is tricking me, but I recall updating the CAIP-25 spec to have this be mandatory, maybe i'm thinking about something else

playground/browser-playground/src/App.tsx
wenfix: 86d93786

packages/connect-multichain/src/multichain/transports/default/index.ts
adonesky1: I actually think you might be right @ffmcgee725 and the type is wrong

wenfix: f07cdb3c

adonesky1 (dismissed):
LGTM!

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: this should not be empty object. the wallet_createSession call will throw if sessionProperties is defined but empty.

TBH this is a bad API design. I'm not sure why this was specified in the original CAIP-25 specs now.

Please undo this for now

wenfix: 78966fe3

adonesky1 (approved):

jiexi (approved):

PR #143, opened 2026-01-26
Setup NPM Publication for Playground Packages (E2E Testing Enablement)

ffmcgee725:
## Explanation
This PR sets up NPM publication for our browser and react-native playgrounds.

* Change access level from restricted to public in `publishConfig`
* Add `publish:preview` and `since-latest-release` scripts to both react-native and browser playgrounds
* Adds release step script for auto version bump of playgrounds
* Updates `contributing.md`

## Checklist


docs/contributing.md
```diff
@@ -236,7 +236,23 @@ Use the following process to release new packages in this repo:
...
+
+   This script automatically detects which core packages are being released and bumps the patch version of any playground packages that depend on them. It will:
+
+   - Identify packages in `packages/` that have version bumps without corresponding git tags
+   - Find playground packages (`browser-playground`, `react-native-playground`) that depend on those packages
+   - Bump their patch versions and update their changelogs
+
+   Review the changes made by this script before proceeding.
+
+6. **Review and update changelogs.**
```
adonesky1, line 266: [nit] could we create a wrapper script 'release' that would do 
```
create-release-branch -i && yarn bump-playground-versions
```

scripts/bump-dependent-playgrounds.ts
```diff
@@ -0,0 +1,276 @@
...
+    '--no-private',
+    '--json',
+  ]);
+
+  const workspaces = stdout.split('\n').map((line) => JSON.parse(line));
+
+  // Filter to only packages in packages/ directory (not playground/)
+  const corePackages = workspaces.filter(
+    (ws: { location: string }) =>
+      ws.location.startsWith('packages/') && ws.location !== '.',
```
adonesky1: [nit]
```suggestion
      ws.location.startsWith('packages/')
```

second check seems redundant since `ws.location.startsWith('packages/')` would rule out the second case?

scripts/bump-dependent-playgrounds.ts
```diff
@@ -0,0 +1,276 @@
...
+${depsList}
+
+`;
+
+  changelog = changelog.replace(unreleasedPattern, newVersionSection);
+
+  // Update the links at the bottom
+  const unreleasedLinkPattern = new RegExp(
+    `\\[Unreleased\\]: (https://github\\.com/MetaMask/connect-monorepo/compare/${packageName.replace('/', '\\/')}@)[^.]+\\.\\.\\.HEAD`,
+  );
```
adonesky1, line 241: For this regex expression,
`[^.]+` stops at the first dot, so it matches `0` instead of `0.4.0`. Use `[\d.]+` instead.
```suggestion
  const unreleasedLinkPattern = new RegExp(
    `\\[Unreleased\\]: (https://github\\.com/MetaMask/connect-monorepo/compare/${packageName.replace('/', '\\/')}@)[\\d.]+\\.\\.\\.HEAD`,
  );
```

ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/143/commits/b64144e21609c092b9d6158d8c044f2e4419257d

scripts/bump-dependent-playgrounds.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/143/commits/b64144e21609c092b9d6158d8c044f2e4419257d

docs/contributing.md
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/143/commits/b64144e21609c092b9d6158d8c044f2e4419257d

ffmcgee725:
@adonesky1 I also though about adding this extra ci workflow [here](https://github.com/MetaMask/connect-monorepo/pull/143/commits/cdbf40279737f92edc45fc14f035e972c2ef64dd) that makes sure our release PRs are also bumping the playgrounds if any of their dependencies are being bumped up on release. Lmk what you think about this, could be overkill but it seems like an extra safety belt to make sure new devs onboarded into release process follow the guidelines.

scripts/bump-dependent-playgrounds.ts
```diff
@@ -0,0 +1,277 @@
...
+ * whose current version in package.json does NOT have a corresponding git tag.
+ * This indicates the version was bumped by `create-release-branch` and is
+ * awaiting release.
+ *
+ * @returns Array of packages pending release with their names and versions.
+ */
+async function getPackagesToRelease(): Promise<ReleasedPackage[]> {
+  const pendingRelease: ReleasedPackage[] = [];
+
+  // Get all workspace packages (excluding playgrounds and private packages)
```
jiexi:
```suggestion
  // Get all workspace packages (excluding private packages)
```

scripts/bump-dependent-playgrounds.ts
```diff
@@ -0,0 +1,277 @@
...
+ * and bumps its version if so.
+ *
+ * @param playgroundPath - The path to the playground package.
+ * @param packagesToRelease - The list of packages pending release.
+ * @returns True if the playground was bumped, false otherwise.
+ */
+async function bumpPlaygroundIfNeeded(
+  playgroundPath: string,
+  packagesToRelease: ReleasedPackage[],
+): Promise<boolean> {
```
jiexi, line 146: this leaves the possibility that the playground has changes that depend on multiple package changes, but not all of those packages may be being released here

docs/contributing.md
```diff
@@ -236,7 +236,23 @@ Use the following process to release new packages in this repo:
...
+
+   ```
+   yarn bump-playground-versions
+   ```
+
+   This script automatically detects which core packages are being released and bumps the patch version of any playground packages that depend on them. It will:
+
+   - Identify packages in `packages/` that have version bumps without corresponding git tags
+   - Find playground packages (`browser-playground`, `react-native-playground`) that depend on those packages
+   - Bump their patch versions and update their changelogs
```
jiexi, line 260:
```suggestion
   - Bump their patch versions and update their changelogs
     - NOTE: It's possible that a playground has unreleased changes that depend on packages other than the packages being bumped! If this is the case, undo the changes by this tool related to the release of affected playground(s)
   
```

jiexi:
Another awkward thing. This means that if the playground has breaking changes (element id's change, the dapp behaves directly, etc), we have to be careful about it not getting released unless we're ready to consume them in the relevant e2e tests unless we specify exact versions (no ^ ~ ranges) in the package.json for that package in the mobile and extension repos

scripts/bump-dependent-playgrounds.ts
ffmcgee725: That is indeed a complication that comes with the required specs of this task that we want to "automate" playground publishing to NPM when we bump packages it depends on.

Do you have a suggested approach to simplify this issue ?

jiexi (approved):

PR #156, opened 2026-02-03
Fix: QR rejection handling

chakra-guy:
### Problem

When users scan a QR code and reject the connection on MetaMask Mobile, the error was not propagated to the dApp. The connection would just hang instead of showing an error, and the QR modal would remain open.

### Solution

**Error Handling:**
- Handle error responses in `initialConnectionMessageHandler` (for initial QR connections)
- Handle error responses in `handleMessage` (for ongoing requests)
- Use `@metamask/rpc-errors` (`providerErrors.custom()`) for proper EIP-1193 compliant errors
- Preserve wallet error codes when provided, default to `4001` (User Rejected Request)

**Modal Behavior:**
- Close QR modal automatically when user rejects connection (call `unload()` on error)

**Playground Improvements:**
- Display errors for Legacy EVM + Solana connections
- Display errors for Wagmi connections (was only logging to console)

**Code Quality:**
- Refactored `initialConnectionMessageHandler` to use early returns (reduced nesting from 5 to 2 levels)
- Combined ID-based and method-based message matching into single flow
- Removed manual error construction in favor of standard MetaMask error library

### Testing

- Tested QR rejection with Connect Multichain, Legacy EVM, and Wagmi
- All 259 unit tests passing
- Added new unit tests for error handling
- Verified modal closes on rejection
- Verified errors display in UI for all three connection methods

## Checklist

### DIFF
--- a/packages/connect-multichain/src/connect.test.ts
+++ b/packages/connect-multichain/src/connect.test.ts
@@ -485,7 +485,8 @@ function testSuite<T extends MultichainOptions>({
 
       // For web-mobile, timeout might be expected due to deeplink hanging
       if (!timedOut) {
-        t.expect(connectError).toBe(sessionError);
+        // Check error message matches (error object may be wrapped with EIP-1193 code)
+        t.expect(connectError.message).toBe(sessionError.message);
         t.expect(sdk.status === 'disconnected').toBe(true);
       } else {
         // If timed out, at least verify it's not connected
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.test.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.test.ts
@@ -44,16 +44,10 @@ t.describe('MWPTransport', () => {
 
           const requestId = 'test-request-123';
           transport.pendingRequests.set(requestId, {
-            request: {
-              jsonrpc: '2.0',
-              id: requestId,
-              method: 'wallet_createSession',
-            } as any,
-            method: 'wallet_createSession',
             resolve: mockResolve,
             reject: mockReject,
             timeout: mockTimeout,
-          });
+          } as any);
 
           // Simulate error message from mobile wallet (user rejection)
           const errorMessage = {
@@ -103,16 +97,10 @@ t.describe('MWPTransport', () => {
 
           const requestId = 'test-request-456';
           transport.pendingRequests.set(requestId, {
-            request: {
-              jsonrpc: '2.0',
-              id: requestId,
-              method: 'wallet_createSession',
-            } as any,
-            method: 'wallet_createSession',
             resolve: mockResolve,
             reject: mockReject,
             timeout: mockTimeout,
-          });
+          } as any);
 
           // Simulate success message from mobile wallet
           const successMessage = {
@@ -161,16 +149,10 @@ t.describe('MWPTransport', () => {
 
           const requestId = 'test-request-789';
           transport.pendingRequests.set(requestId, {
-            request: {
-              jsonrpc: '2.0',
-              id: requestId,
-              method: 'wallet_createSession',
-            } as any,
-            method: 'wallet_createSession',
             resolve: mockResolve,
             reject: mockReject,
             timeout: mockTimeout,
-          });
+          } as any);
 
           // Error without message field (invalid format per EIP-1193)
           const errorMessage = {
@@ -208,16 +190,10 @@ t.describe('MWPTransport', () => {
 
         const requestId = 'test-request-custom';
         transport.pendingRequests.set(requestId, {
-          request: {
-            jsonrpc: '2.0',
-            id: requestId,
-            method: 'wallet_createSession',
-          } as any,
-          method: 'wallet_createSession',
           resolve: mockResolve,
           reject: mockReject,
           timeout: mockTimeout,
-        });
+        } as any);
 
         // Error with custom code (e.g., 4100 = Unauthorized)
         const errorMessage = {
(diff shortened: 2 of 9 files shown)


packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -149,9 +150,30 @@ export class MWPTransport implements ExtendedTransport {
...
+            if ('error' in messagePayload && messagePayload.error) {
+              this.pendingRequests.delete(messagePayload.id);
+              const errorData = messagePayload.error as {
+                message?: string;
+                code?: number;
+              };
+              request.reject(
+                providerErrors.custom({
+                  code: errorData.code ?? 4001,
+                  message: errorData.message ?? 'Request rejected by user',
```
jiexi: not sure about these default values.

wow. so were we missing request.reject from errors this whole time? I guess that makes sense because I feel like the DefaultTransport has handling errors differently than the MWPTransport. sigh, good catch

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -391,38 +411,51 @@ export class MWPTransport implements ExtendedTransport {
...
+              // Handle error response (e.g., user rejected the connection)
+              if (messagePayload.error) {
+                const errorData = messagePayload.error as {
+                  message?: string;
+                  code?: number;
+                };
+                return rejectConnection(
+                  providerErrors.custom({
+                    code: errorData.code ?? 4001,
+                    message: errorData.message ?? 'Connection rejected',
```
jiexi: not sure about these defaults

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -149,9 +150,30 @@ export class MWPTransport implements ExtendedTransport {
     if (typeof message === 'object' && message !== null) {
       if ('data' in message) {
         const messagePayload = message.data as Record<string, unknown>;
+
         if ('id' in messagePayload && typeof messagePayload.id === 'string') {
           const request = this.pendingRequests.get(messagePayload.id);
+
           if (request) {
+            clearTimeout(request.timeout);
+
```
ffmcgee725, line 180: Thank you for the clean up refactor on `handleMessage` 🙏🏾

packages/connect-multichain/src/multichain/transports/mwp/index.ts
ffmcgee725: I'm okay with the message. But what would you suggest as the default value for `code` then ?

packages/connect-multichain/src/connect.test.ts
```diff
@@ -485,7 +485,8 @@ function testSuite<T extends MultichainOptions>({
 
       // For web-mobile, timeout might be expected due to deeplink hanging
       if (!timedOut) {
-        t.expect(connectError).toBe(sessionError);
+        // Check error message matches (error object may be wrapped with EIP-1193 code)
```
ffmcgee725, line 488: [nit] okay with removing this comment

ffmcgee725 (dismissed):

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -149,9 +150,30 @@ export class MWPTransport implements ExtendedTransport {
...
+              const errorData = messagePayload.error as {
+                message?: string;
+                code?: number;
+              };
+              request.reject(
+                providerErrors.custom({
+                  code: errorData.code ?? 4001,
+                  message: errorData.message ?? 'Request rejected by user',
+                }),
+              );
```
adonesky1:
```suggestion
       const errorData = messagePayload.error as Record<string, unknown>;
       
              if (
                typeof errorData.code === 'number' &&
                typeof errorData.message === 'string'
              ) {
                request.reject(
                  providerErrors.custom({
                    code: errorData.code,
                    message: errorData.message,
                  }),
                );
              } else {
                // Non-standard error format - preserve original error info
                request.reject(
                     rpcErrors.internal({
                      message: `Unexpected error format: ${JSON.stringify(messagePayload.error)}`,
                    }),
                );
              }
```

packages/connect-multichain/src/multichain/transports/mwp/index.ts
adonesky1: Possibly this: https://github.com/MetaMask/connect-monorepo/pull/156/changes#r2766075398

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: this works, maybe drop the `Unexpected error format: `?

Since we need this in two places, might worth putting into a helper.

Thank you 🙏

adonesky1 (commented):
Just [a request](https://github.com/MetaMask/connect-monorepo/pull/156/changes#r2766075398) to improve error serialization but otherwise LGTM!

packages/connect-multichain/src/multichain/transports/mwp/index.test.ts
```diff
@@ -0,0 +1,250 @@
...
+          const mockTimeout = setTimeout(() => {}, 60000) as any;
+
+          const requestId = 'test-request-123';
+          transport.pendingRequests.set(requestId, {
+            request: {
+              jsonrpc: '2.0',
+              id: requestId,
+              method: 'wallet_createSession',
+            } as any,
+            method: 'wallet_createSession',
```
jiexi: not needed

jiexi: should only be resolve, reject, and timeout. same for below

jiexi (approved):

PR #157, opened 2026-02-03
feat: ConnectMultichain Singleton

jiexi:
## Explanation

This PR introduces a singleton pattern for the Multichain SDK that shifts the source of truth for connection status for downstream ecosystem clients (ConnectEVM, ConnectSolana) to the `wallet_sessionChanged` event.

1. Adds a singleton for the actual MetaMaskConnectMultichain instance itself with logic to merge incoming constructor options with the existing constructor options of the singleton

2. Removes the `disconnect()` on `connect()` pattern

3. Calling `connect()` while there is an inflight MWP connection results in a `'Existing connection is pending. Please check your MetaMask Mobile app to continue.'` error being thrown
    - This mirrors the existing Default transport behavior in which multiple permission requests cannot be inflight at once (request of type `wallet_requestPermissions` already pending for origin...)

4. `connect()` is now additive and will combine the requested scopes and accounts with the existing (if any)
    - Also had to remove `wallet_revokeSession` call in `DefaultTransport.connect()`

5. `disconnect()` now supports partial revokes. If some scopes still remain after a partial revoke, the underlying connection remains intact and can be reused.

6. Fix/Adjust Browser Playground
    - Adds disconnect buttons to Browser playground cards
    - Fixes EVM Provider setChainId and setAccounts to also use the `connect` event
    - Removes the explicit ActiveProviderStorage pattern. Now all providers (cards) are "Active" even without user input to connect to a specific ecosystem
    - Multichain Card and button properly reflect initial instantiation connecting status

7. Makes `wallet_sessionChanged` the source of truth for `ConnectEvm` and future ecosystem clients

Addresses: https://consensyssoftware.atlassian.net/browse/WAPI-1066

## Checklist


adonesky1:
https://github.com/user-attachments/assets/54dde6cf-645a-4e57-8dd0-8f99eb4ad87c

## EDIT

After testing further I see that even if my initial connection is with the legacy connector I see that same error 🤔 

I'm starting to wonder whether we should even have this error "Requests cannot be made to chains not explicitly configured in supportedNetworks."

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -821,20 +852,54 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
+      ),
+    );
 
-    this.#listener = undefined;
-    this.#beforeUnloadListener = undefined;
-    this.#transport = undefined;
-    this.#providerTransportWrapper.clearNotificationCallbacks();
-    this.#dappClient = undefined;
+    // in theory this is only needed for MWP
+    this.emit('wallet_sessionChanged', { sessionScopes: newSessionScopes });
```
jiexi: investigate if this can be moved into MWP transport

packages/connect-multichain/src/domain/multichain/index.ts
```diff
@@ -70,9 +74,64 @@ export abstract class MultichainCore extends EventEmitter<SDKEvents> {
...
   }
+
+  /**
+   * Merges the given options into the current instance options.
+   * Only the mergeable keys are updated (api.supportedNetworks, ui.*, mobile.*, transport.extensionId, debug).
+   * Used when createMultichainClient is called with an existing singleton.
+   *
+   * @param partial - Options to merge/overwrite onto the current instance
+   */
+  mergeOptions(partial: MergeableMultichainOptions): void {
```
jiexi, line 92: spec this

packages/connect-evm/src/connect.ts
```diff
@@ -126,9 +129,30 @@ export class MetamaskConnectEVM {
      *
      * @param session - The session data
      */
-    this.#sessionChangedHandler = (session): void => {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    this.#sessionChangedHandler = async (session): Promise<void> => {
```
jiexi: spec this

packages/connect-evm/src/connect.ts
```diff
@@ -485,17 +480,21 @@ export class MetamaskConnectEVM {
...
     logger('request: disconnect');
 
-    await this.#core.disconnect();
+    const sessionScopes = this.#sessionScopes;
+    const eip155Scopes = Object.keys(sessionScopes).filter((scope) => {
+      const { namespace } = parseScopeString(scope as Scope);
+      return namespace === 'eip155';
+    });
+
+    await this.#core.disconnect(eip155Scopes as Scope[]);
```
jiexi, line 466: spec this

packages/connect-evm/src/connect.ts
```diff
@@ -334,64 +358,35 @@ export class MetamaskConnectEVM {
...
-          logger('transport-event: chainChanged', notificationChainId);
-          // Cache the chainId for persistence across page refreshes
-          this.#cacheChainId(notificationChainId).catch((error) => {
-            logger('Error caching chainId in notification handler', error);
+      // Wait for the wallet_sessionChanged event to fire and set the provider properties
+      return new Promise((resolve) => {
+        this.#provider.once('connect', ({ chainId, accounts }) => {
+          logger('fulfilled-request: connect', {
+            chainId,
+            accounts,
```
jiexi, line 345: hard one to spec, but let's try

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -258,10 +254,17 @@ export class DefaultTransport implements ExtendedTransport {
     });
   }
 
-  async disconnect(): Promise<void> {
-    this.#notificationCallbacks.clear();
+  async disconnect(scopes: Scope[] = []): Promise<void> {
```
jiexi, line 257: spec this

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -511,22 +511,60 @@ export class MWPTransport implements ExtendedTransport {
...
+    this.kvstore.set(
+      SESSION_STORE_KEY,
+      JSON.stringify({
+        result: {
+          sessionScopes: newSessionScopes,
+        },
+      }),
+    );
+
+    // TODO: update chain_store too. Emit chainChanged
```
jiexi: TODO

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -725,18 +741,43 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
+      new Set([...existingCaipChainIds, ...scopes]),
+    ) as Scope[];
+    const requestedCaipAccountIds = Array.from(
+      new Set([...existingCaipAccountIds, ...caipAccountIds]),
+    ) as CaipAccountId[];
+    const requestedSessionProperites = {
+      ...sessionData.sessionProperties,
+      ...sessionProperties,
+    };
+
```
jiexi: add a helper for merging

playground/browser-playground/src/sdk/SDKProvider.tsx
```diff
@@ -61,25 +55,18 @@ export const SDKProvider = ({ children }: { children: React.ReactNode }) => {
...
-                setSession(payload.params as SessionData);
-              }
-            } else if (payload.method === 'stateChanged') {
-              setStatus(payload.params as ConnectionStatus);
-            }
-          },
         },
       });
+
+      // TODO: Check if we can get rid of transport.onNotification constructor param
```
jiexi, line 61: TODO

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/91e31b0fd5988df4e3ae2f2aea8d86c7cc0b4587

playground/browser-playground/src/sdk/SDKProvider.tsx
jiexi: maybe too much to bite off in this PR

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/e52031166b2edab98f152c2728db4cf33509f039

packages/connect-evm/src/connect.ts
```diff
@@ -126,9 +129,30 @@ export class MetamaskConnectEVM {
      *
      * @param session - The session data
      */
-    this.#sessionChangedHandler = (session): void => {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
adonesky1, line 134: is this still necessary?

packages/connect-evm/src/connect.ts
```diff
@@ -334,64 +358,35 @@ export class MetamaskConnectEVM {
...
+            accounts,
+            chainId: chainId as Hex,
+          });
+        });
+      });
+    } catch (error) {
+      logger('Error connecting to wallet', error);
+      throw error;
+    } finally {
+      this.#status = 'disconnected';
```
adonesky1: This seems incorrect? Seems like it should only happen in the `catch` block?

packages/connect-evm/src/connect.ts
```diff
@@ -485,17 +480,21 @@ export class MetamaskConnectEVM {
...
+    const eip155Scopes = Object.keys(sessionScopes).filter((scope) => {
+      const { namespace } = parseScopeString(scope as Scope);
+      return namespace === 'eip155';
+    });
+
+    await this.#core.disconnect(eip155Scopes as Scope[]);
     this.#onDisconnect();
     this.#clearConnectionState();
 
     this.#core.off('wallet_sessionChanged', this.#sessionChangedHandler);
```
adonesky1: Not related to your changes but we are not removing the same function object we have bound to our listener. Not sure if we should resolve separately or just here?

packages/connect-evm/src/connect.ts
jiexi: this is broken. don't need `bind(this)`

packages/connect-evm/src/connect.ts
```diff
@@ -334,64 +358,35 @@ export class MetamaskConnectEVM {
...
-        if (notification?.method === 'metamask_chainChanged') {
-          // @ts-expect-error TODO: address this
-          const notificationChainId = notification?.params?.chainId;
-          logger('transport-event: chainChanged', notificationChainId);
-          // Cache the chainId for persistence across page refreshes
-          this.#cacheChainId(notificationChainId).catch((error) => {
-            logger('Error caching chainId in notification handler', error);
+      // Wait for the wallet_sessionChanged event to fire and set the provider properties
+      return new Promise((resolve) => {
+        this.#provider.once('connect', ({ chainId, accounts }) => {
```
adonesky1, line 342: should we be cleaning up this listener in the catch?

packages/connect-evm/src/connect.ts
```diff
@@ -126,9 +129,30 @@ export class MetamaskConnectEVM {
...
+          { result: string[]; id: number; jsonrpc: '2.0' }
+        >({ method: 'eth_accounts', params: [] });
+
+        const chainId = await this.#getSelectedChainId(hexPermittedChainIds);
+
+        this.#onConnect({
+          chainId,
+          accounts: initialAccounts.result as Address[],
+        });
+      }
```
adonesky1:
```suggestion
      const hexPermittedChainIds = getPermittedEthChainIds(this.#sessionScopes);
      if (hexPermittedChainIds.length === 0) {
        this.#onDisconnect();
      } else {
     
        const initialAccounts = await this.#core.transport.sendEip1193Message<
          { method: 'eth_accounts'; params: [] },
          { result: string[]; id: number; jsonrpc: '2.0' }
        >({ method: 'eth_accounts', params: [] });

        const chainId = await this.#getSelectedChainId(hexPermittedChainIds);

        this.#onConnect({
          chainId,
          accounts: initialAccounts.result as Address[],
        });
      }
```

packages/connect-evm/src/connect.ts
```diff
@@ -102,6 +102,9 @@ export class MetamaskConnectEVM {
   /** The clean-up function for the notification handler */
   #removeNotificationHandler?: () => void;
 
+  /** The current connection status */
+  #status: 'disconnected' | 'connected' | 'connecting' = 'disconnected';
```
ffmcgee725: I know we already use this pattern in many places across the codebase, but would now be a good time to start thinking about using an `enum` instead ?

packages/connect-evm/src/connect.ts
```diff
@@ -126,9 +129,27 @@ export class MetamaskConnectEVM {
...
+          { result: string[]; id: number; jsonrpc: '2.0' }
+        >({ method: 'eth_accounts', params: [] });
+
+        const chainId = await this.#getSelectedChainId(hexPermittedChainIds);
+
+        this.#onConnect({
+          chainId,
+          accounts: initialAccounts.result as Address[],
+        });
+      }
```
ffmcgee725:
```suggestion
       const initialAccounts = await this.#core.transport.sendEip1193Message<
          { method: 'eth_accounts'; params: [] },
          { result: Address[]; id: number; jsonrpc: '2.0' }
        >({ method: 'eth_accounts', params: [] });

        const chainId = await this.#getSelectedChainId(hexPermittedChainIds);

        this.#onConnect({
          chainId,
          accounts: initialAccounts.result,
        });
      }
```

no need for typecasting this way

packages/connect-evm/src/connect.ts
ffmcgee725: Yes, I believe so as well

packages/connect-multichain/src/domain/multichain/index.ts
```diff
@@ -70,9 +74,64 @@ export abstract class MultichainCore extends EventEmitter<SDKEvents> {
...
+      if (partial.ui.preferExtension !== undefined) {
+        uiUpdates.preferExtension = partial.ui.preferExtension;
+      }
+      if (partial.ui.showInstallModal !== undefined) {
+        uiUpdates.showInstallModal = partial.ui.showInstallModal;
+      }
+      if (Object.keys(uiUpdates).length > 0) {
+        opts.ui = { ...opts.ui, ...uiUpdates };
+      }
+    }
```
ffmcgee725: I feel like optional chaining `headless`, `preferExtension`, and `showInstallModal` would look a bit cleaner here, as it removes the need for the extra `if` condition check on `ui !== undefined`, since there's plenty of `if` in this code block already.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -511,22 +512,73 @@ export class MWPTransport implements ExtendedTransport {
...
-      typeof window !== 'undefined' &&
-      typeof window.removeEventListener !== 'undefined' &&
-      this.windowFocusHandler
-    ) {
-      window.removeEventListener('focus', this.windowFocusHandler);
-      this.windowFocusHandler = undefined;
+  async disconnect(scopes: Scope[] = []): Promise<void> {
+    const cachedSession = await this.getCachedResponse({
+      jsonrpc: '2.0',
+      id: '0',
```
ffmcgee725, line 523: is cached response `id` guaranteed to be `0` ?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -869,4 +926,19 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
+      const response = await this.transport.request({
+        method: 'wallet_getSession',
+      });
+      if (response.result) {
+        this.emit('wallet_sessionChanged', response.result);
+      } else {
+        this.emit('wallet_sessionChanged', { sessionScopes: {} });
+      }
+    }
+  }
```
ffmcgee725, line 963:
```suggestion
  async emitSessionChanged(): Promise<void> {
    const emptySession = { sessionScopes: {} };

    if (this.status !== 'connected' && this.status !== 'connecting') {
      this.emit('wallet_sessionChanged', emptySession);
      return;
    }

    const response = await this.transport.request({
      method: 'wallet_getSession',
    });

    this.emit('wallet_sessionChanged', response.result ?? emptySession);
}
```

Less nesting, wdyt ?

playground/browser-playground/src/sdk/SDKProvider.tsx
ffmcgee725: Even the above changes to the UI on the playgrounds, feels like they don't belong in this PR (Although I see the need to have them in order to properly test stuff out)

Should we perhaps either
- Extract these playground changes onto separate granular PR
- Update this task/ticket description to also fit this criteria in, that way will be easier to track in the future if necessary where the changes were made

adonesky1: I would favor 

> Update this task/ticket description to also fit this criteria in, that way will be easier to track in the future if necessary where the changes were made

since, as you said these changes are required to properly test these changes even for reviewers

packages/connect-evm/src/connect.ts
adonesky1: Lets leave this to followup cleanup. Not sure if there is an advantage to it here in any case?

packages/connect-evm/src/connect.ts
adonesky1: https://github.com/MetaMask/connect-monorepo/pull/157/commits/fe8233e75f9fc6a66878343160543e52404604ec

packages/connect-evm/src/connect.ts
adonesky1: resolved here https://github.com/MetaMask/connect-monorepo/pull/170

packages/connect-evm/src/connect.ts
adonesky1: https://github.com/MetaMask/connect-monorepo/pull/157/commits/f0b569b66e6e2c66713acea92a2c6bd400876277

packages/connect-evm/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/f930bf7420862b03dd0590a889d006cfadd295ef

adonesky1: https://github.com/MetaMask/connect-monorepo/pull/157/changes/f930bf7420862b03dd0590a889d006cfadd295ef

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -260,25 +286,16 @@ export class MetaMaskConnectMultichain extends MultichainCore {
 
   async #init(): Promise<void> {
```
chakra-guy, line 299: 1. Can we document clearly which options are intended to be "mergeable" vs "first caller wins"? For example, `mergeOptions` intentionally does not merge `storage`, `ui.factory`, or `transport.onNotification`. Should we warn (or throw) if a later caller passes different values for non-mergeable fields?

2. What should happen if the first `instancePromise` rejects during `#init()`? Right now the rejected promise stays cached on the global object, so future `create()` calls will keep failing until a full page refresh, no?

packages/connect-multichain/src/multichain/index.ts
chakra-guy, line 709: One thing I'm worried about is concurrent `connect()` calls (e.g. EVM and Solana wrappers calling connect around the same time).

In particular, `connect()` starts with:
- `if (this.status !== 'connected') { await this.disconnect(); }`

Questions:
- If two callers call `connect()` concurrently while status is `connecting`, does the second call intentionally cancel/tear down the first? Or do we want to serialize connects (store an "in-flight connect promise" and have other callers await it)?
- Do we want to explicitly guard against calling `connect()` while `status === 'connecting'` to avoid flakiness / weird intermediate states?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: the `id` here doesn't matter. It's just to make typescript happy. `getCachedResponse` only matches against the `method`

packages/connect-multichain/src/multichain/index.ts
jiexi: Great point. The reason why this `disconnect()` call exists was to deal with the scenario where the user starts a deeplink connect but does not complete it before returning to the dapp. I'm forgetting the specifics on why we chose to do it this way rather than simple deeplink the user back in to the wallet (which I think was the previous behavior) if they press connect again before the timeout.

@adonesky1 we need to think this through again

packages/connect-evm/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/f45468c45293d2834b471b3bda206d2560410d56

packages/connect-multichain/src/multichain/index.ts
jiexi: 1. Yeah, all great points. Will add documentation. A warning would be good. I don't think we should throw.
2. It shouldn't be possible for `#init()`to reject currently, but I've added some logic in `create()` to address your comment. https://github.com/MetaMask/connect-monorepo/pull/157/commits/1426333b36f64d152f022ec9adb7816dd686fd0c

packages/connect-multichain/src/domain/multichain/index.ts
jiexi: Very fair. Something like this perhaps? https://github.com/MetaMask/connect-monorepo/pull/157/commits/89e45a68a558bad4debfcaec3fde7f066f59fd3b

packages/connect-multichain/src/multichain/index.ts
jiexi: 1. Actually re: warning. We're going to have console.warn noise from our own clients. In theory I imagine developers will not call `MultichainClient.create()` directly, but rather use our `createMultichainClient` helpers which would provide the values for `storage` and `ui.factory`. I was also planning on getting rid of `transport.onNotification` as an option since I think it's possible to get the same events by listening to the `MultichainClient` itself, but I have not fully verified that yet. If all of the above is true, perhaps we don't have a need to throwa. warning?

packages/connect-multichain/src/domain/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/d9a97351e86dff5ddb1226db7880af123d91053f

packages/connect-evm/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/e21639393ebbece99a5c794a01dc2474aa47deff

packages/connect-multichain/src/domain/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/d9a97351e86dff5ddb1226db7880af123d91053f

packages/connect-evm/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/ec18eb4528b333e0988ce581b7bc31c40ce7f047

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: skipping

packages/connect-evm/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/0b0ae780d9cc6ea20177767b734f5dbeaa781934

packages/connect-multichain/src/multichain/index.ts
jiexi: Alex D and I settled on reopening the existing deeplink and throwing a "please check your wallet app" error if the connection status is `connecting` and the transport is `MWP`. Extension currently disallows concurrent permission requests and will throw immediately, and so it makes sense for force Mobile into the same pattern for now. It seems to make more sense to throw instead of add a promise guard because it's not clear to me how to resolve the scenario where there are multiple clients (i.e. more than 1) waiting on a connection to complete at the same time

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -486,9 +518,11 @@ export class MWPTransport implements ExtendedTransport {
         );
       }
 
+      const storedSessionRequest = await this.getStoredSessionRequest();
+
       timeout = setTimeout(() => {
         reject(new TransportTimeoutError());
-      }, this.options.connectionTimeout);
+      }, storedSessionRequest ? this.options.resumeTimeout : this.options.connectionTimeout);
```
jiexi: TODO: this needs to move into a PR based off this branch

adonesky1:
Comments/questions just on the PR description:
> Remove wallet_revokeSession call in DefaultTransport.connect(), making it an additive action

Not sure what this means? Can you spell it out a bit more?

> Make wallet_sessionChanged the source of truth for ConnectEvm

And all compatibility layers right?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -795,7 +836,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         this.#deeplinkConnect(
           scopes,
           caipAccountIds,
```
adonesky1: right? why wouldn't we be using the merged scopes for this flow as we are for the others above?

```suggestion
requestedScopes,
requestedCaipAccountIds,
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -724,19 +752,32 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       logger('Error tracking connection_initiated event', error);
     }
 
+    const sessionData = await this.#getCaipSession();
+
+    const {
+      requestedScopes,
+      requestedCaipAccountIds,
+      requestedSessionProperties,
```
adonesky1: rename for clarity?
```suggestion
      mergedScopes,
      mergedCaipAccountIds,
      mergedSessionProperties,
```

packages/connect-multichain/src/multichain/transports/multichainApiClientWrapper/index.ts
```diff
@@ -53,16 +56,24 @@ export class MultichainApiClientWrapperTransport implements Transport {
     });
   }
 
-  setupNotifcationListener(): void {
-    this.metamaskConnectMultichain.transport.onNotification(
-      this.notifyCallbacks.bind(this),
-    );
+  clearTransportNotifcationListener(): void {
```
adonesky1:
```suggestion
  clearTransportNotificationListener(): void {
```

packages/connect-multichain/src/multichain/transports/multichainApiClientWrapper/index.ts
```diff
@@ -53,16 +56,24 @@ export class MultichainApiClientWrapperTransport implements Transport {
...
+    this.notificationListener?.();
+    this.notificationListener = undefined;
+  }
+
+  setupTransportNotifcationListener(): void {
```
adonesky1:
```suggestion
  setupTransportNotificationListener(): void {
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -155,16 +163,40 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
-    );
-    if (isEnabled) {
-      enableDebug('metamask-sdk:core');
+    const globalObject = getGlobalObject();
+    const existing = globalObject[SINGLETON_KEY] as
+      | Promise<MetaMaskConnectMultichain>
+      | undefined;
+    if (existing) {
+      const instance = await existing;
+      instance.mergeOptions(options);
```
adonesky1, line 178: Should we call out which options are mutable versus those that are fixed by the first client who initiatlized?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -511,22 +511,80 @@ export class MWPTransport implements ExtendedTransport {
...
+      },
+    );
+
+    // Clear the cached values for eth_accounts and eth_chainId if all eip155 scopes were removed.
+    const remainingScopesIncludeEip155 = remainingScopes.some((scope) =>
+      scope.includes('eip155'),
+    );
+    if (!remainingScopesIncludeEip155) {
+      this.kvstore.delete(ACCOUNTS_STORE_KEY);
+      this.kvstore.delete(CHAIN_STORE_KEY);
```
adonesky1, line 557: seems strange that there is evm specific code at this layer?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -215,7 +247,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         if (hasExtensionInstalled) {
           const apiTransport = new DefaultTransport();
           this.#transport = apiTransport;
-          this.#providerTransportWrapper.setupNotifcationListener();
+          this.#providerTransportWrapper.setupTransportNotifcationListener();
```
adonesky1:
```suggestion
          this.#providerTransportWrapper.setupTransportNotificationListener();
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -227,7 +259,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         const apiTransport = new MWPTransport(dappClient, kvstore);
         this.#dappClient = dappClient;
         this.#transport = apiTransport;
-        this.#providerTransportWrapper.setupNotifcationListener();
+        this.#providerTransportWrapper.setupTransportNotifcationListener();
```
adonesky1:
```suggestion
        this.#providerTransportWrapper.setupTransportNotificationListener();
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -868,4 +936,19 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
+      this.emit('wallet_sessionChanged', emptySession);
+      return;
+    }
+
+    const response = await this.transport.request({
+      method: 'wallet_getSession',
+    });
+
+    this.emit('wallet_sessionChanged', response.result ?? emptySession);
+  }
```
adonesky1, line 963: perhaps worth spelling out some of what's going on here with inline comments. Who is listening to this and why can't they just listen directly on the transport?

packages/connect-multichain/src/multichain/transports/multichainApiClientWrapper/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/7a74583c8d3b21b0291dffe88a1907b6f9930e4a

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/7a74583c8d3b21b0291dffe88a1907b6f9930e4a

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/7a74583c8d3b21b0291dffe88a1907b6f9930e4a

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: Right, but caching is MWP specific. Unfortunately I think it would be more awkward to have this in the ConnectEvm layer than it is here in the MwpTransport

packages/connect-multichain/src/multichain/transports/multichainApiClientWrapper/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/f3ba9ceeb32ab75d7782f0f3180413d147f7071a

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/abb2d689cfbadf850a680a0f8390287b95a05edd

packages/connect-multichain/src/multichain/index.ts
jiexi: yeah, missed this. Proof that existing additive UI layer logic is working lol

jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/64a78f8fc1f395cc0a681ff159c8063617c7f41b

jiexi:
PR desc updated

packages/connect-evm/src/connect.ts
```diff
@@ -334,64 +332,36 @@ export class MetamaskConnectEVM {
...
+            chainId: chainId as Hex,
+          });
+        });
+      });
 
-    logger('fulfilled-request: connect', {
-      chainId: chainIds[0],
-      accounts: this.#provider.accounts,
-    });
+      await this.#core.connect(
```
wenfix, line 354: [nit/tech-debt]:  make connect parameters an object  for readability and avoiding passing `undefined`

packages/connect-evm/src/connect.ts
```diff
@@ -748,6 +722,34 @@ export class MetamaskConnectEVM {
...
+    logger('event: wallet_sessionChanged', session);
+    this.#sessionScopes = session?.sessionScopes ?? {};
+    const hexPermittedChainIds = getPermittedEthChainIds(this.#sessionScopes);
+    if (hexPermittedChainIds.length === 0) {
+      this.#onDisconnect();
+    } else {
+      let initialAccounts: Address[] = [];
+      if (this.#core.status === 'connected') {
+        const ethAccountsResponse =
+          await this.#core.transport.sendEip1193Message<
```
wenfix: [nit/tech-debt]: extract commonly used methods and returns into types to avoid these verbose type parameters

packages/connect-evm/src/connect.ts
ffmcgee725: Another small observation regarding this

in the `connect.test.ts` we have

```ts
type Status =
  | 'connected'
  | 'disconnected'
  | 'connecting'
  | 'loaded'
  | 'pending';
```

Which is not aligned with this one, which is missing `loaded` and `pending`

packages/connect-evm/src/connect.ts
```diff
@@ -795,8 +803,40 @@ export class MetamaskConnectEVM {
...
-    this.#provider.emit('connect', data);
-    this.#eventHandlers?.connect?.(data);
+    if (this.#status !== 'connected') {
+      this.#status = 'connected';
+      this.#provider.emit('connect', data);
+      this.#eventHandlers?.connect?.(data);
+
+      this.#removeNotificationHandler?.();
+
+      // TODO: Verify if #core.on('metamask_accountsChanged') and #core.on('metamask_chainChanged')
```
ffmcgee725, line 815: Should we either create a ticket for this TODO, verify it, or remove it ?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -155,16 +163,40 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
-      enableDebug('metamask-sdk:core');
+    const globalObject = getGlobalObject();
+    const existing = globalObject[SINGLETON_KEY] as
+      | Promise<MetaMaskConnectMultichain>
+      | undefined;
+    if (existing) {
+      const instance = await existing;
+      instance.mergeOptions(options);
+      if (options.debug) {
+        enableDebug('metamask-sdk:*');
```
wenfix, line 180: just for my understanding, why are going back to the `-sdk` nomenclature?

playground/browser-playground/src/components/LegacyEVMCard.tsx
```diff
@@ -173,6 +175,13 @@ export function LegacyEVMCard({
         <h3 data-testid={TEST_IDS.legacyEvm.title} className="text-lg font-semibold text-gray-800 truncate">
           Legacy EVM Connection
         </h3>
+        <button
+          type="button"
+          onClick={disconnect}
+          className="bg-red-500 text-white px-3 py-1 rounded text-sm hover:bg-red-600 transition-colors"
+        >
+          Disconnect
+        </button>
```
ffmcgee725, line 184: Needs a `data-testid` for e2e test interactions

playground/browser-playground/src/components/SolanaWalletCard.tsx
```diff
@@ -115,10 +112,21 @@ export const SolanaWalletCard: React.FC = () => {
...
+          Solana Wallet
+        </h3>
+        {connected && (
+          <button
+            type="button"
+            onClick={disconnect}
+            className="bg-red-500 text-white px-3 py-1 rounded text-sm hover:bg-red-600 transition-colors"
+          >
+            Disconnect
+          </button>
```
ffmcgee725, line 127: Needs a `data-testid` for e2e test interactions

playground/browser-playground/src/components/WagmiCard.tsx
```diff
@@ -60,6 +62,13 @@ export function WagmiCard() {
         <h3 data-testid={TEST_IDS.wagmi.title} className="text-lg font-semibold text-gray-800 truncate">
           Wagmi Connection
         </h3>
+        <button
+          type="button"
+          onClick={() => disconnect()}
+          className="bg-red-500 text-white px-3 py-1 rounded text-sm hover:bg-red-600 transition-colors"
+        >
+          Disconnect
+        </button>
```
ffmcgee725, line 71: Needs a `data-testid` for e2e test interactions

wenfix (dismissed):

packages/connect-multichain/src/multichain/transports/multichainApiClientWrapper/index.ts
```diff
@@ -168,8 +176,13 @@ export class MultichainApiClientWrapperTransport implements Transport {
...
     }
 
+    const revokeSessionParams = request.params as
+      | RevokeSessionParams<RPCAPI>
+      | undefined;
+    const scopes = revokeSessionParams?.scopes ?? [];
+
     try {
-      this.metamaskConnectMultichain.disconnect();
+      this.metamaskConnectMultichain.disconnect(scopes as Scope[]);
```
chakra-guy: `disconnect()` is async but isn't awaited here. This means `{ result: true }` is returned to the caller before the disconnect has actually completed; the `wallet_getSession` check, transport teardown, and status change all happen after the response is sent.

In practice: any code that does `await provider.revokeSession()` and then checks connection state or navigates will find the core still in `connected` status.

```suggestion
      await this.metamaskConnectMultichain.disconnect(scopes as Scope[]);
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -683,8 +706,13 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     sessionProperties?: SessionProperties,
     forceRequest?: boolean,
   ): Promise<void> {
-    if (this.status !== 'connected') {
-      await this.disconnect();
+    if (
```
chakra-guy, line 715: Following up on our earlier thread about concurrent connect() calls. I understand the decision to throw for MWP and let the extension's own rejection handle the browser case, rather than adding a promise guard. 

The concern @jiexi raised about "how to resolve N clients waiting" is fair.

But after testing this with the singleton in place, the practical outcome might not be ideal. When an app auto-connects both EVM and Solana on load, whichever connect() fires second gets this raw rejection from the extension:

> Request of type 'wallet_requestPermissions' already pending for origin ...

The caller has no way to distinguish "you should retry after the other connect finishes" from a genuine failure. So it surfaces as a silent partial-connect: only one chain works, and the app doesn't know why.

I think the "N clients waiting" concern is actually solvable here: the mutex doesn't need to manage multiple parallel sessions, it just needs to queue them. 

But I might be overstating here the size of this bug and it might be a super super edge case..

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -724,19 +752,32 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       logger('Error tracking connection_initiated event', error);
     }
 
+    const sessionData = await this.#getCaipSession();
+
+    const {
+      mergedScopes,
+      mergedCaipAccountIds,
+      mergedSessionProperties,
+    } = mergeRequestedSessionWithExisting(
```
chakra-guy: The scope merge logic here correctly computes the union of scope IDs. But when these merged IDs reach `DefaultTransport.connect()`, `getOptionalScopes()` rebuilds them as `{ methods: [], notifications: [], accounts: [] }` for every scope, including ones that already exist in the wallet session with real method grants.

The resulting `wallet_createSession` request looks like:
```
optionalScopes: {
  'eip155:1': { methods: [], notifications: [], accounts: [] },  // was granted with real methods
  'solana:...': { methods: [], notifications: [], accounts: [] }  // new scope
}
```

The wallet appears to treat `wallet_createSession` as a full session replacement rather than an additive merge, so the existing EVM scope gets re-requested with empty capabilities. 

Possible fix: when building `createSessionParams` in `DefaultTransport.connect()`, preserve the existing `sessionScopes` objects (methods, notifications, accounts) for scopes that are already granted, and only use `getOptionalScopes()` for genuinely new scopes:

```ts
const existingScopes = walletSession?.sessionScopes ?? {};
const optionalScopes = {};
for (const scope of options?.scopes ?? []) {
  optionalScopes[scope] = existingScopes[scope] ?? { methods: [], notifications: [], accounts: [] };
}
```

playground/browser-playground/src/components/LegacyEVMCard.tsx
jiexi: Absolutely. I have this on child branch here https://github.com/MetaMask/connect-monorepo/pull/174. Tried to keep this PR more concise

playground/browser-playground/src/components/SolanaWalletCard.tsx
jiexi: see above https://github.com/MetaMask/connect-monorepo/pull/174

packages/connect-evm/src/connect.ts
jiexi: Absolutely. There's actually a `TODO` comment above the method implementation already 😭 🙏

jiexi: Perhaps we ticket this

packages/connect-multichain/src/multichain/index.ts
jiexi: mistake

jiexi: looks like there are several places in the codebase were we use the `metamask-sdk` prefix for logging

packages/connect-multichain/src/multichain/index.ts
jiexi: Good catch. So this isn't actually an issue until we add capability buckets to `wallet_createSession` (e.g. granting `personal_sign` also gives you access to all other signing methods). Currently `wallet_createSession` effectively grants all methods regardless of input `methods` value. Because of this, it's currently okay to pass `[]` for the methods value even though that will not match the actual methods reflected in the returned result from `wallet_createSession` and `wallet_sessionChanged`.

This is also awkward because currently `ConnectMultichain.connect()` is responsible for merging the scopes and caipAccountIds values before passing them to `#transport.connect()`. We either need to change `Transport.connect()` to accept a full `wallet_createSession` param object OR we need to moving merging of scopes and caipAccountIds into the Transport itself. 

I think for now we leave this as is.

packages/connect-multichain/src/multichain/index.ts
jiexi: In practice since we've released Solana with the injected provider, I don't believe we have seen dapps actually try to simultaneous connect both the EVM and Solana provider. If they use both and want to connect both, they always do use sequentially. 

> only one chain works, and the app doesn't know why.

I mean the dapp has to parse our `already pending for origin` error to know what to do. Not all errors are not standardized and dapps already have to check several possible ones when working with different wallets.

> it just needs to queue them.

I thought about this approach too. In general getting a connection approval prompt that you did not initiate yourself is jarring. Getting back to back connection approval prompts you did not initiate would be even more jarring. I don't think we should deviate from the injected provider behavior at this time, but definitely will keep this option in the back of mind.

packages/connect-evm/src/connect.ts
jiexi: yeah... definitely should overload the sendEip1193Message typedef for these commons ones. Maybe worth a tech debt ticket

packages/connect-evm/src/connect.ts
jiexi: good catch. https://github.com/MetaMask/connect-monorepo/pull/157/commits/7441fbc20f30a86b3814a87a2d6ba0140aeb6b6f

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/aa366a55791eb97c391dc4cb28b1e97daf27b298

packages/connect-multichain/src/multichain/transports/mwp/index.ts
adonesky1: 🤔 not sure if I agree. Let me think on it a bit more and get back here

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/157/commits/d31249d3fc5ce3e27c01201ab11bc3943781242d

packages/connect-evm/src/connect.ts
jiexi: let's not do that for now. We don't have too many internal calls of sendEip1193Message. Cleaned up this type templating by just typecasting the final result to what we expect https://github.com/MetaMask/connect-monorepo/pull/157/commits/5d4b50a7c6518ffc1973d0eb2737c94cdae5a52a

packages/connect-multichain/src/multichain/index.ts
jiexi: This is in progress already. Thank you for the callout!

packages/connect-multichain/src/multichain/index.ts
jiexi: Will be adding a ticket to track doing this maybe in the future

adonesky1 (approved):
LGTM! Let's go!!! Amazing work Jiexi 🚀 🚀 🚀

PR #170, opened 2026-02-06
fix: preserve instance-scoped event listeners on disconnect and fix beforeunload listener leak

chakra-guy:
## Summary

- **`display_uri` / `wallet_sessionChanged` not firing on reconnect**: `MetamaskConnectEVM.disconnect()` was removing the `display_uri` and `wallet_sessionChanged` listeners from the core instance, but `connect()` never re-registered them. After a disconnect→reconnect cycle (especially in headless mode), the QR code URI was never emitted and session scope updates were silently lost. Fix: stop removing these instance-scoped listeners on disconnect — they should live for the lifetime of the SDK instance. Session-scoped listeners (notification handler) are still correctly cleaned up.

- **`beforeunload` listener never actually removed (memory leak)**: `#createBeforeUnloadListener()` called `.bind(this)` separately in `addEventListener` and `removeEventListener`, producing two different function references. Since `removeEventListener` requires the exact same reference, the cleanup was silently a no-op, leaking a listener on every connect/disconnect cycle. Fix: store the bound function once and reuse it.

## Checklist

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix `display_uri` and `wallet_sessionChanged` events not firing after disconnect and reconnect in headless mode ([#170](https://github.com/MetaMask/connect-monorepo/pull/170))
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


wenfix (dismissed):

ffmcgee725 (approved):

adonesky1 (approved):

adonesky1:
Great catches 🙇

PR #176, opened 2026-02-18
feat: re-prompt with connect deeplink when attempting to connect while an existing connection is in-progress 

jiexi:
## Explanation

When `ConnectMultichain.connect()` is called while a connection is already pending, the user is re-prompted with the pending connection deeplink

## Checklist


ffmcgee725 (dismissed):

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -124,6 +133,27 @@ export class MWPTransport implements ExtendedTransport {
     }
   }
 
+  private async removeStoredSessionRequest(): Promise<void> {
```
adonesky1: Worth being hyperspecific about what this is doing? Because its already a difficult flow to understand the what and why of
```suggestion
  private async removeStoredPendingSessionRequest(): Promise<void> {
```

adonesky1: Also maybe worth adding a jsdoc or comment about why we store the pending session request at all and why we have to be vigilant about removing the stored value?

adonesky1 (approved):
LGTM!

PR #183, opened 2026-02-23
fix: Fix JSON-RPC id generation

jiexi:
## Explanation

Fix bug with JSON-RPC requests where previously used `id` value could be re-used causing the wallet to ignore the request when a dapp was refreshed/reloaded
## References

Fixes: https://consensyssoftware.atlassian.net/browse/WAPI-1120

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/transports/default/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/default/index.ts
@@ -14,6 +14,7 @@ import type { ExtendedTransport, RPCAPI, Scope, SessionData } from 'src/domain';
 import {
   addValidAccounts,
   getOptionalScopes,
+  getUniqueRequestId,
   getValidAccounts,
   isSameScopesAndAccounts,
 } from '../../utils';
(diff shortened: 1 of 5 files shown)


packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -148,8 +146,7 @@ export class DefaultTransport implements ExtendedTransport {
     this.#setupMessageListener();
 
     // Generate unique request ID - increment counter to ensure uniqueness
-    this.#reqId += 1;
-    const requestId = `${this.#reqId}`;
+    const requestId = String(getUniqueRequestId());
```
adonesky1, line 149: So was this the place where the duplication was occurring? Since it was still using datebased generation?

adonesky1 (dismissed):
LGTM

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: MWP was starting the request Id at 0 was the reason

adonesky1 (approved):

PR #187, opened 2026-02-25
fix: Fix React Native Playground dapp url

jiexi:
## Explanation

Fixes an issue where the react native playground was unable to get Solana signMessage prompts to appear on the wallet. This was caused by an value being used for the dapp url param which fails `new URL()` on the wallet side. This is fixed by using a better url value. 

Separately, we need to investigate how we should implement better handling for malformed dapp urls.

## Checklist


adonesky1 (dismissed):
LGTM, amazing digging here Jiexi 🙇

adonesky1:
Do we need to change it here too? https://github.com/MetaMask/connect-monorepo/blob/524edeb711e53b83cb31a6cdee4f43a3cf894dda/playground/react-native-playground/src/wagmi/metamask-connector.ts#L340

ffmcgee725 (commented):
@adonesky1 not there directly, but in the wagmi-connector that lives under `integrations/wagmi/metamask-connector.ts`, and then those changes are copied over to the playgrounds that need a wagmi integration.

jiexi:
let's just leave the window.location.href polyfill in for now

adonesky1 (dismissed):
LGTM

adonesky1 (approved):

PR #192, opened 2026-03-02
ci: add workflow to build and publish React Native Playground APK on release

ffmcgee725:
## Explanation

Adds automated CI infrastructure to build the React Native Playground Android APK and attach it to GitHub Releases, eliminating the manual build-and-transfer process that blocks running multichain RN connect E2E tests in metamask-mobile.

#### What

- **New workflow** (`.github/workflows/build-rn-playground-apk.yml`):
  - `workflow_call` (reusable from the release pipeline) + `workflow_dispatch` (manual/ad-hoc builds from any branch)
  - Full self-contained build: checkout → Node/Yarn/JDK setup → `yarn build` → Infura `.env` from secret → `expo prebuild` → `gradlew assembleRelease`
  - Uploads APK as a workflow artifact (90-day retention)
  - On release: attaches versioned `rn-playground-<version>.apk` to the GitHub Release via `gh release upload`
  - Gradle and Yarn caching for faster subsequent builds

- **Release pipeline integration** (`publish-release.yml`, `main.yml`):
  - New `build-rn-playground-apk` job runs after `get-release-version` (parallel with the existing gh-pages deployments)
  - Forwards `INFURA_API_KEY` secret through the workflow call chain

#### Why

The metamask-mobile Appwright E2E tests (multichain RN connect specs) require a prebuilt playground APK. Until now, generating it required manually cloning connect-monorepo, building locally, and transferring the APK — a process that doesn't scale for CI or new contributors. With this change, every release automatically produces a fetchable APK artifact.

#### Setup required

Add the `INFURA_API_KEY` secret to the connect-monorepo repository settings. The APK is signed with Expo's default debug keystore, which is sufficient for a test dApp.

### Files changed

| File | Change |
|------|--------|
| `.github/workflows/build-rn-playground-apk.yml` | New reusable + dispatchable workflow |
| `.github/workflows/publish-release.yml` | New `build-rn-playground-apk` job, `INFURA_API_KEY` secret passthrough |
| `.github/workflows/main.yml` | Forward `INFURA_API_KEY` to `publish-release.yml` |
| `playground/react-native-playground/README.md` | CI build docs, secrets table, BrowserStack investigation |

## Checklist

### DIFF
--- a/.github/workflows/build-rn-playground-apk.yml
+++ b/.github/workflows/build-rn-playground-apk.yml
@@ -3,8 +3,8 @@ name: Build React Native Playground APK
 on:
   workflow_call:
     inputs:
-      release_version:
-        description: 'Release version to use for APK naming and release asset upload (e.g., "17.0.0")'
+      release_tag:
+        description: 'Git tag of the GitHub Release to attach the APK to (e.g., "v17.0.0")'
         required: false
         type: string
       upload_to_release:
@@ -17,8 +17,8 @@ on:
         required: false
   workflow_dispatch:
     inputs:
-      release_version:
-        description: 'Version label for the APK (leave empty to use package.json version)'
+      release_tag:
+        description: 'Git tag to attach the APK to (leave empty to skip release upload)'
         required: false
         type: string
 
@@ -33,16 +33,11 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ github.sha }}
-          fetch-depth: 0
 
       - name: Determine APK version
         id: version
         run: |
-          if [ -n "${{ inputs.release_version }}" ]; then
-            VERSION="${{ inputs.release_version }}"
-          else
-            VERSION=$(node -p "require('./playground/react-native-playground/package.json').version")
-          fi
+          VERSION=$(node -p "require('./playground/react-native-playground/package.json').version")
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
           echo "APK version: $VERSION"
 
@@ -91,8 +86,6 @@ jobs:
           if [ -n "$INFURA_KEY" ]; then
             echo "EXPO_PUBLIC_INFURA_API_KEY=$INFURA_KEY" > .env
             echo "Infura API key configured from secret"
-          elif [ -f .env ]; then
-            echo "Using existing .env file"
           else
             cp .env.example .env
             echo "Warning: No INFURA_API_KEY secret set; using empty .env"
@@ -131,15 +124,14 @@ jobs:
           retention-days: 90
 
       - name: Upload APK to GitHub Release
-        if: inputs.upload_to_release == true && inputs.release_version != ''
+        if: inputs.upload_to_release == true && inputs.release_tag != ''
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          TAG=$(git describe --tags --abbrev=0 2>/dev/null || echo "v${{ inputs.release_version }}")
-          echo "Uploading ${{ steps.apk.outputs.apk_name }} to release $TAG..."
+          echo "Uploading ${{ steps.apk.outputs.apk_name }} to release ${{ inputs.release_tag }}..."
 
-          gh release upload "$TAG" \
+          gh release upload "${{ inputs.release_tag }}" \
             "${{ steps.apk.outputs.apk_name }}" \
             --clobber
 
-          echo "APK uploaded to release $TAG"
+          echo "APK uploaded to release ${{ inputs.release_tag }}"
--- a/.github/workflows/build-rn-playground-apk.yml
+++ b/.github/workflows/build-rn-playground-apk.yml
@@ -1,0 +1,145 @@
+name: Build React Native Playground APK
+
+on:
+  workflow_call:
+    inputs:
+      release_version:
+        description: 'Release version to use for APK naming and release asset upload (e.g., "17.0.0")'
+        required: false
+        type: string
+      upload_to_release:
+        description: 'Whether to upload the APK as a GitHub Release asset'
+        required: false
+        type: boolean
+        default: false
+    secrets:
+      INFURA_API_KEY:
+        required: false
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: 'Version label for the APK (leave empty to use package.json version)'
+        required: false
+        type: string
+
+jobs:
+  build-rn-playground-apk:
+    name: Build Android APK
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.sha }}
+          fetch-depth: 0
+
+      - name: Determine APK version
+        id: version
+        run: |
+          if [ -n "${{ inputs.release_version }}" ]; then
+            VERSION="${{ inputs.release_version }}"
+          else
+            VERSION=$(node -p "require('./playground/react-native-playground/package.json').version")
+          fi
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "APK version: $VERSION"
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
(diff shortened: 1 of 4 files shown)


playground/react-native-playground/README.md
```diff
@@ -169,18 +182,101 @@ See [scripts/README.md](./scripts/README.md) for detailed polyfill documentation
...
+   Adding them to connect-monorepo would expand the secret surface area.
+
+2. **Consistency** — The existing metamask-mobile pattern (build/download APK →
+   upload to BrowserStack → run tests) already works for the MetaMask wallet
+   APK via Bitrise. The playground APK follows the same pattern, just with
+   GitHub Releases as the source instead of Bitrise.
+
+3. **Freshness control** — BrowserStack apps expire after 30 days of inactivity.
+   Uploading per-test-run guarantees the APK is always available regardless of
+   connect-monorepo release cadence.
```
jiexi, line 261: TIL

.github/workflows/build-rn-playground-apk.yml
```diff
@@ -0,0 +1,145 @@
...
+      - name: Configure Infura API key
+        working-directory: playground/react-native-playground
+        env:
+          INFURA_KEY: ${{ secrets.INFURA_API_KEY }}
+        run: |
+          if [ -n "$INFURA_KEY" ]; then
+            echo "EXPO_PUBLIC_INFURA_API_KEY=$INFURA_KEY" > .env
+            echo "Infura API key configured from secret"
+          elif [ -f .env ]; then
+            echo "Using existing .env file"
```
jiexi: i'm having difficulty figuring out where in the CI steps `.env` is being created besides here

.github/workflows/publish-release.yml
```diff
@@ -141,3 +143,15 @@ jobs:
       destination_dir: latest
     secrets:
       PUBLISH_DOCS_TOKEN: ${{ secrets.PUBLISH_DOCS_TOKEN }}
+
+  build-rn-playground-apk:
```
jiexi, line 174: would this trigger on release of any package? Do we need to make this step run only if the react native playground is specifically being released?

jiexi: perhaps we need to target a new tag entirely, not the root level package.json version?

adonesky1: This is a good point. I think I had pointed Joao down this path of targeting any release, but I suppose it wouldn't make sense to run this unless the react-native-playground itself is being released

adonesky1: I think as @jiexi suggests we need a new tag format specifically for the rn playground package to do the proper gating comparison

.github/workflows/publish-release.yml
```diff
@@ -141,3 +143,15 @@ jobs:
...
+    name: Build and publish RN Playground APK
+    needs: get-release-version
+    permissions:
+      contents: write
+    uses: ./.github/workflows/build-rn-playground-apk.yml
+    with:
+      release_version: ${{ needs.get-release-version.outputs.RELEASE_VERSION }}
```
adonesky1: Per the comments above, this should probably be the react native playground version rather than the monorepo version

.github/workflows/build-rn-playground-apk.yml
adonesky1: yeah I think this is dead code for CI? And I don't think we'll ever run this locally?

.github/workflows/build-rn-playground-apk.yml
```diff
@@ -0,0 +1,145 @@
...
+          name: rn-playground-apk
+          path: ${{ steps.apk.outputs.apk_name }}
+          retention-days: 90
+
+      - name: Upload APK to GitHub Release
+        if: inputs.upload_to_release == true && inputs.release_version != ''
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAG=$(git describe --tags --abbrev=0 2>/dev/null || echo "v${{ inputs.release_version }}")
```
adonesky1: `git describe --tags --abbrev=0` might resolve to a stale or unrelated tag (e.g., a tag from a different package in the monorepo). Since inputs.release_version is already available, this should just use `v${{ inputs.release_version }}` directly to guarantee the right release is targeted.

.github/workflows/build-rn-playground-apk.yml
ffmcgee725: On .env (line 94-95): That elif is actually dead code in CI. After a fresh actions/checkout, there's no .env on disk -- it will always hit either the secret branch or the .env.example fallback. The branch only has value if someone runs the workflow locally with act or similar and happens to have a .env present. I'll remove it to avoid confusion.

.github/workflows/publish-release.yml
ffmcgee725: Yes, it would run on every monorepo release regardless of which package changed. Since the playground depends on workspace packages (`@metamask/connect-multichain`, `@metamask/connect-evm`, etc.), rebuilding the APK on any release is actually desirable -- the playground bundles those packages at build time, so a new `@metamask/connect-multichain` release means the APK should be rebuilt to include it. The APK filename will still reflect the playground's own version, which only bumps when the playground itself changes (enforced by `validate-release-playground-bumps.yml`).

.github/workflows/build-rn-playground-apk.yml
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/192/commits/3408204687f4b70049e9fb0ec7b86750058925ff

.github/workflows/publish-release.yml
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/192/commits/3408204687f4b70049e9fb0ec7b86750058925ff

.github/workflows/publish-release.yml
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/192/commits/3408204687f4b70049e9fb0ec7b86750058925ff

jiexi: > means the APK should be rebuilt to include it.

Not necessarily though. The playground may contain breaking changes that the E2E tests have not been updated for yet

jiexi: I think that on the mobile side when we specify the github release url (identified by tag) to download the built RN playground APK from, that APK should never change. If the APK changes, it should be only if the github release url (i.e. tag) changes too. 

If we allow clobbering the existing apk in a tag, this violates the assumptions of versioning (since there is a version in the github release url)

ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/192/commits/b4ea0d65676d8071da4a0f7c1877c35a9b608e28

adonesky1: I don’t think https://github.com/MetaMask/connect-monorepo/commit/b4ea0d65676d8071da4a0f7c1877c35a9b608e28 fully closes the loop on Jiexi’s concern yet. The new gate is checking whether the RN playground tag already exists at all, not whether the playground was actually released in this specific run, so it could still fire on unrelated releases. On top of that, the APK is still being uploaded to the generic `RELEASE_TAG`, and with `--clobber` we can still overwrite an asset behind an existing release URL, which is the versioning issue Jiexi called out.

I think the safer path is:
1. only run the APK job when the RN playground itself is part of the current release
2. upload the APK to the RN-playground-specific tag (for example `@metamask/react-native-playground@x.y.z`), not the generic release tag
 3. avoid `--clobber` once that tag-to-APK mapping is versioned and stable

That way the mobile side can rely on a release URL never changing underneath it.

ffmcgee725: addressed concern here
https://github.com/MetaMask/connect-monorepo/pull/192/commits/c74d8813ebd3401fd043a009dbfc927cc01ea937

adonesky1: This looks right to me now! Will let @jiexi confirm though!

adonesky1 (approved):
LGTM

PR #198, opened 2026-03-05
feat: passive wallet_sessionChanged listening for extension

jiexi:
## Explanation

Uses the DefaultTransport on init if no existing transport is stored, extension is found, and extensionIsPreferred. Ensures that wallet_sessionChanged events are listened to and MultichainClient.status is correctly updated.

To test, use `window.ethereum.request({method: 'eth_requestAccounts})` in console to trigger the permission flow outside of the MMC lifecycle 

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -258,7 +258,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     if (transportType) {
       if (transportType === TransportType.Browser) {
         if (hasExtensionInstalled) {
-          // Why does this not use setupDefaultTransport?..
           const apiTransport = new DefaultTransport();
           this.#transport = apiTransport;
           this.#providerTransportWrapper.setupTransportNotificationListener();
@@ -268,7 +267,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
           return apiTransport;
         }
       } else if (transportType === TransportType.MWP) {
-        // Why does this not use setupMWP?..
         const { adapter: kvstore } = this.options.storage;
         const dappClient = await this.#createDappClient();
         const apiTransport = new MWPTransport(dappClient, kvstore);
@@ -281,7 +279,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         return apiTransport;
       }
 
-      await this.storage.removeTransport(); // why do we remove the transport here?..
+      await this.storage.removeTransport();
     }
 
     return undefined;
@@ -295,7 +293,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         await this.transport.connect();
       }
       this.status = 'connected';
-      // Why do do we this?..
       if (this.transport instanceof MWPTransport) {
         await this.storage.setTransport(TransportType.MWP);
       } else {
@@ -307,7 +304,8 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       if (hasExtensionInstalled && preferExtension) {
         await this.#setupDefaultTransport();
         // We don't actually want getStoredTransport to return this transport
-        // since we aren't connecting to it
+        // since we aren't connecting it to the wallet with a CAIP session, only
+        // listening for wallet_sessionChanged events.
         await this.storage.removeTransport();
         // Normally calling DefaultTransport.connect() ensures that the transport is initialized
         // and that wallet_sessionChanged (faked) is emitted. But because we are not
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -313,7 +313,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         // and that wallet_sessionChanged (faked) is emitted. But because we are not
         // calling transport.connect(), we need to initialize DefaultTransport manually.
         await this.transport.init();
-        await this.emitSessionChanged();
       }
       this.status = 'loaded';
     }
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -96,6 +96,9 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   }
 
   set status(value: ConnectionStatus) {
+    if (this._status === value) {
+      return;
+    }
     this._status = value;
     this.options.transport?.onNotification?.({
       method: 'stateChanged',
@@ -243,6 +246,12 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         const sessionScopes =
           (payload.params as SessionData | undefined)?.sessionScopes ?? {};
         const hasScopes = Object.keys(sessionScopes).length > 0;
+        // We don't want to set the status to disconnected if this is the first wallet_sessionChanged event
+        // triggered during #setupTransport() initialization for passive extension wallet_sessionChanged listening.
+        if (this.status === 'pending' && !hasScopes) {
+          this.status = 'loaded';
+          return;
+        }
         this.status = hasScopes ? 'connected' : 'disconnected';
       }
 
@@ -301,6 +310,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     } else {
       const hasExtensionInstalled = await hasExtension();
       const preferExtension = this.options.ui.preferExtension ?? true;
+      // Setup passive listening for extension wallet_sessionChanged events
       if (hasExtensionInstalled && preferExtension) {
         await this.#setupDefaultTransport();
         // We don't actually want getStoredTransport to return this transport
@@ -316,7 +326,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
           // Passive init may fail if extension transport isn't ready
         }
       }
-      this.status = 'loaded';
     }
   }
 
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -261,6 +261,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     if (transportType) {
       if (transportType === TransportType.Browser) {
         if (hasExtensionInstalled) {
+          // Why does this not use setupDefaultTransport?..
           const apiTransport = new DefaultTransport();
           this.#transport = apiTransport;
           this.#providerTransportWrapper.setupTransportNotificationListener();
@@ -270,6 +271,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
           return apiTransport;
         }
       } else if (transportType === TransportType.MWP) {
+          // Why does this not use setupMWP?..
         const { adapter: kvstore } = this.options.storage;
         const dappClient = await this.#createDappClient();
         const apiTransport = new MWPTransport(dappClient, kvstore);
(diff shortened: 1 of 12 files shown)


packages/connect-multichain/src/multichain/index.ts
```diff
@@ -239,6 +239,16 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
     ) {
+      if (payload.method === 'wallet_sessionChanged') {
+        const sessionScopes = (payload.params as SessionData | undefined)
+          ?.sessionScopes ?? {};
+        const hasScopes = Object.keys(sessionScopes).length > 0;
+        this.status =
+          hasScopes
+            ? 'connected'
+            : 'disconnected';
+      }
```
jiexi, line 255: needed due to passive listening via DefaultTransport

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -272,8 +282,16 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         return apiTransport;
       }
 
-      await this.storage.removeTransport();
-    }
+      await this.storage.removeTransport(); // why do we remove the transport here?..
```
jiexi: odd...

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -184,16 +185,45 @@ export class DefaultTransport implements ExtendedTransport {
     });
   }
 
+  async init(): Promise<void> {
+    this.#setupMessageListener();
+    await this.#transport.connect();
+  }
+
+  async destroy(): Promise<void> {
```
jiexi: this is never called. and we never intend it to be called, i think? So should we just remove it entirely (and get rid of the `#listeners` too since we don't need to remove their event listeners?)

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -246,6 +276,9 @@ export class DefaultTransport implements ExtendedTransport {
       }
       walletSession = response.result as SessionData;
     }
+    // I see... this is needed because we don't have listeners setup before we connect.
+    // but even if we had listeners setup, we may have missed the session changed event already.
+    // so we fake it here.
```
jiexi: This begs the question. Should this be coming from the Transport or the MultichainClient?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -251,6 +261,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     if (transportType) {
       if (transportType === TransportType.Browser) {
         if (hasExtensionInstalled) {
+          // Why does this not use setupDefaultTransport?..
```
jiexi: problem for another PR i suppose

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -260,6 +271,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
           return apiTransport;
         }
       } else if (transportType === TransportType.MWP) {
+          // Why does this not use setupMWP?..
```
jiexi: problem for another PR i suppose

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -286,12 +298,27 @@ export class MetaMaskConnectMultichain extends MultichainCore {
         await this.transport.connect();
       }
       this.status = 'connected';
+      // Why do do we this?..
       if (this.transport instanceof MWPTransport) {
         await this.storage.setTransport(TransportType.MWP);
       } else {
         await this.storage.setTransport(TransportType.Browser);
       }
```
jiexi: we literally call getStoredTransport right before this. It's already set. Why are we setting it again?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -983,7 +1017,8 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   async emitSessionChanged(): Promise<void> {
     const emptySession = { sessionScopes: {} };
 
-    if (this.status !== 'connected' && this.status !== 'connecting') {
+    // Unsure if this change is okay for MWP. May need to adjust or undo
+    if (!this.transport?.isConnected()) {
```
jiexi: seems like this is fine?.. at least for DefaultTransport. Need to validate for MWP

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/198/commits/3845e071db8c3b3c6cf347145078811a0ee875cb

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/198/commits/67da8e1c5f64d4374c2c02ac5c2d133a64850224

packages/connect-multichain/src/multichain/index.ts
jiexi: https://consensyssoftware.atlassian.net/browse/WAPI-1153

packages/connect-multichain/src/multichain/index.ts
jiexi: https://consensyssoftware.atlassian.net/browse/WAPI-1153

packages/connect-multichain/src/multichain/index.ts
jiexi: https://consensyssoftware.atlassian.net/browse/WAPI-1153

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/198/commits/075bb6bff43520633c28dba327f4b0e8b8bbc57e

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -246,46 +267,10 @@ export class DefaultTransport implements ExtendedTransport {
       }
       walletSession = response.result as SessionData;
     }
-    this.#notifyCallbacks({
-      method: 'wallet_sessionChanged',
-      params: walletSession,
-    });
```
adonesky1, line 252: This will not happen anymore in the refactor since its only happening in the public `init` and not the private one. This doesn't seem intentional?

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -246,46 +267,10 @@ export class DefaultTransport implements ExtendedTransport {
...
-      window.removeEventListener('message', this.#handleNotificationListener);
-      this.#handleNotificationListener = undefined;
-    }
-
-    // Reject all pending requests
-    for (const [, request] of this.#pendingRequests) {
-      clearTimeout(request.timeout);
-      request.reject(new Error('Transport disconnected'));
-    }
-    this.#pendingRequests.clear();
```
adonesky1, line 286: Help me understand why we don't need any of this cleanup anymore?

packages/connect-multichain/src/multichain/transports/default/index.ts
```diff
@@ -184,16 +190,31 @@ export class DefaultTransport implements ExtendedTransport {
...
+    await this.#init();
+    let walletSession: SessionData = { sessionScopes: {} };
+    try {
+      const sessionRequest = await this.request(
+        { method: 'wallet_getSession' },
+        this.#defaultRequestOptions,
+      );
+      walletSession = sessionRequest.result as SessionData;
+    } catch {
+      // wallet_getSession may fail if extension is not ready; treat as no session
```
adonesky1: should add a logger here?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -292,6 +299,23 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
       }
     } else {
+      const hasExtensionInstalled = await hasExtension();
+      const preferExtension = this.options.ui.preferExtension ?? true;
+      if (hasExtensionInstalled && preferExtension) {
+        await this.#setupDefaultTransport();
+        // We don't actually want getStoredTransport to return this transport
+        // since we aren't connecting it to the wallet with a CAIP session, only
+        // listening for wallet_sessionChanged events.
+        await this.storage.removeTransport();
```
adonesky1: nit: Alternatively we could add a `passive` parameter to `#setupDefaultTransport` to skip the storage write?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -292,6 +299,23 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
+        // Normally calling DefaultTransport.connect() ensures that the transport is initialized
+        // and that wallet_sessionChanged (faked) is emitted. But because we are not
+        // calling transport.connect(), we need to initialize DefaultTransport manually.
+        try {
+          await this.transport.init();
+        } catch {
+          // Passive init may fail if extension transport isn't ready
+        }
+      }
       this.status = 'loaded';
```
adonesky1: Heads up — there's a status flicker here. When init() fires, it emits `wallet_sessionChanged` synchronously through `#notifyCallbacks`, which hits `#onTransportNotification` and sets this.status to 'connected' or 'disconnected' (depending on scopes). Then this line immediately overwrites it with 'loaded'. Since the status setter emits `stateChanged` to consumers on every write, they'll see a quick connected → loaded (or disconnected → loaded) flash — two state change notifications in rapid succession where the first one is misleading.

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: If the DefaultTransport is instantiated, there is no longer any scenario in which we want to remove the listeners / actually fully disconnect. It's currently not possible to switch between DefaultTransport and MwpTransport without disabling extension AND reloading the page

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/198/commits/7f180a40bae56b37feff79fb8493294185b9d0b9

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/198/commits/07a0199996e4b82c606d8d39a46047255ab38574

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/198/commits/a545b901b424fc9b8eb38eec988536753797fd9c

packages/connect-multichain/src/multichain/transports/default/index.ts
jiexi: typed a bunch before realizing that i do indeed have a gap in logic here. Thank you for this call out. https://github.com/MetaMask/connect-monorepo/pull/198/commits/7cf128e32319abe0bedd5a13652b30b52eb1150c

adonesky1 (approved):
This probably should have a changelog entry for connect-multichain? Otherwise LGTM

PR #204, opened 2026-03-10
fix: Fix broken wallet_sessionChanged listener when using underlying multichain-api-client's WindowPostMessageTransport

jiexi:
## Explanation

Fixes wallet_sessionChanged events not making it to the MultichainClient from the DefaultTransport because the underlying [WindowPostMessageTransport](https://github.com/MetaMask/multichain-api-client/blob/decd68f641188720c81a7000dbcbcf1ae6bcbda5/src/transports/windowPostMessageTransport.ts#L91) has removed any event listeners when `connect()` is called on it and it is already connected. Should it be doing that, i'm not sure. 

The workaround for now is to check if we are connected or not in the  (multichain-api-client) transport used by the (our) DefaultTransport (not a good name...) before attempting to call connect() again.

## Checklist

### DIFF
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+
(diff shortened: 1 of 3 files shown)


adonesky1 (approved):
LGTM

PR #205, opened 2026-03-11
fix(connect-evm): connect method always returns default chain id

wenfix:
## Explanation

Even when chain ids were specified, the `connect` method returned the default chain id. It's consumers, `connectWith` and `connectAndSign`, were also incorrectly using the default chain id for their purposes. This PR addresses that. Also adds a button to `browser-playground` to enable using `connectAndSign` for testing. 

## References

Fixes [WAPI-1233](https://consensyssoftware.atlassian.net/browse/WAPI-1233)

## Steps 

1. `yarn && yarn build`
2. Start a local browser-playground instance (`yarn workspace @metamask/browser-playground start`)
3. Select Linea
4. Press `Connect And Sign (Legacy EVM)`
5. Confirm connection
6. The `personal_sign` prompt should be for Linea

## Checklist

### DIFF
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -108,6 +108,14 @@ export class MetamaskConnectEVM {
   #status: ConnectEvmStatus = 'disconnected';
 
   /**
+   * The preferred chain ID to use for the active connect() call.
+   * Set to the first explicit chainId passed by the caller, cleared once the
+   * connect event resolves. Takes priority over the cache in #getSelectedChainId
+   * so that an explicit chainIds request always wins over a prior cached value.
+   */
+  #pendingPreferredChainId: Hex | undefined;
+
+  /**
    * Creates a new MetamaskConnectEVM instance.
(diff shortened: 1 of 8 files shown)


packages/connect-evm/src/connect.ts
```diff
@@ -358,11 +384,15 @@ export class MetamaskConnectEVM {
...
-      return result as Promise<{ accounts: Address[]; chainId: Hex }>;
+      // Await here so the finally block always runs after #onSessionChanged
+      // has consumed #pendingPreferredChainId, not before.
+      return await result;
     } catch (error) {
       this.#status = 'disconnected';
       logger('Error connecting to wallet', error);
       throw error;
+    } finally {
+      this.#pendingPreferredChainId = undefined;
```
jiexi, line 395: need to verify that there isn't a race condition here where this line runs before the wallet_sessionChanged handler runs. I think there is, just let me double check

adonesky1 (approved):

PR #206, opened 2026-03-11
feat: change mmconnect_version from string to versioned object

adonesky1:
## Summary

Changes the `mmconnect_version` analytics property from a plain version string (`"0.0.0"`) to a `Record<string, string>` object keyed by connect package name, e.g.:

```json
{
  "connect-multichain": "0.8.0",
  "connect-evm": "0.6.0"
}
```

Only packages that are actually loaded appear in the object — `connect-multichain` is always present since it's the core singleton, while `connect-evm` and `connect-solana` only appear when their modules are imported by the consumer.

**Jira:** [WAPI-1246](https://consensyssoftware.atlassian.net/browse/WAPI-1246) &  https://consensyssoftware.atlassian.net/browse/WAPI-1232

**Companion PR (analytics API backend):** [consensys-vertical-apps/metamask-sdk-analytics-api#24](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/24)

### New: Version Registry (`connect-multichain/src/domain/versions.ts`)
- `ConnectPackageName` type union: `'connect-multichain' | 'connect-evm' | 'connect-solana'`
- `ConnectVersions` type: `connect-multichain` required, others optional
- `registerPackageVersion(name, version)` — called at module-load time by each package
- `getVersions()` — returns snapshot of all currently registered versions
- `connect-multichain` self-registers using its own `package.json` version

### Package Registration
- **connect-evm/src/index.ts** — registers `connect-evm` version at module load via `import pkg from '../package.json'`
- **connect-solana/src/index.ts** — registers `connect-solana` version at module load via `import pkg from '../package.json'`
- Versions come directly from each package's `package.json` — no build-time injection, works in all environments (build, test, dev)

### Analytics Wiring
- **multichain/utils/analytics.ts** — `getBaseAnalyticsProperties` and `getWalletActionAnalyticsProperties` now return `mmconnect_version: ConnectVersions` via `getVersions()` instead of the old `getVersion()` string
- **multichain/index.ts** — `#setupAnalytics()` sets the global property with the versions object; `getVersion()` retained for non-analytics uses (SDK info string, modals)

### Schema
- **analytics/src/schema.ts** — `mmconnect_version` type changed from `string` to `Record<string, string>`

### Tests
- **analytics.test.ts** — updated test properties to use version object
- **init.test.ts** — assertion changed from `expect.any(String)` to `expect.any(Object)`

## Deployment Order

> **The companion analytics API PR must be deployed first** to avoid validation failures. The backend needs to accept the object format before the SDK starts sending it.

1. Deploy analytics API change first ([metamask-sdk-analytics-api#24](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/24))
2. Then merge and release this SDK change

## Test Plan

- [ ] Verify `connect-multichain` version is always present in analytics events
- [ ] Verify `connect-evm` version appears only when `@metamask/connect-evm` is imported
- [ ] Verify `connect-solana` version appears only when `@metamask/connect-solana` is imported
- [ ] Run existing unit tests: `yarn test` in connect-multichain, connect-evm, analytics packages
- [ ] Coordinate deployment with companion analytics API PR


packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -46,13 +47,12 @@ export async function getBaseAnalyticsProperties(
   options: MultichainOptions,
   storage: StoreClient,
 ): Promise<{
-  mmconnect_version: string;
+  mmconnect_version: ConnectVersions | Partial<ConnectVersions>;
```
jiexi: not sure if both these types should be here. I'm assuming plain `ConnectVersions` is correct

packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -82,22 +82,21 @@ export async function getWalletActionAnalyticsProperties(
   storage: StoreClient,
   invokeOptions: InvokeMethodOptions,
 ): Promise<{
-  mmconnect_version: string;
+  mmconnect_version: ConnectVersions | Partial<ConnectVersions>;
```
jiexi: same

packages/connect-solana/tsup.config.ts
```diff
@@ -9,6 +9,10 @@ const peerDeps = Object.keys((pkg as any).peerDependencies ?? {});
 const external = [...deps, ...peerDeps];
 const entryName = (pkg as any).name.replace('@metamask/', '');
 
+const versionDefine = {
+  __PACKAGE_VERSION__: JSON.stringify((pkg as any).version),
```
jiexi: why does this one need a `as any` but the other tsup don't?

jiexi (dismissed):

ffmcgee725 (dismissed):

wenfix (approved):

PR #211, opened 2026-03-11
feat: change `getInfuraRpcUrls` to accept param object and chain ids property

jiexi:
## @metamask/connect-multichain
- getInfuraRpcUrls(infuraApiKey) → getInfuraRpcUrls({ infuraApiKey, caipChainIds? })
- When caipChainIds is provided, only matching CAIP-2 chains are included in the output
## @metamask/connect-evm
- getInfuraRpcUrls(infuraApiKey) → getInfuraRpcUrls({ infuraApiKey, chainIds? })
- chainIds accepts hex values (e.g., ['0x1', '0x89']), converts to CAIP format internally, and delegates to the multichain helper

## References

Fixes: https://consensyssoftware.atlassian.net/browse/WAPI-1234

## Checklist

### DIFF
--- a/packages/connect-evm/src/utils/infura.ts
+++ b/packages/connect-evm/src/utils/infura.ts
@@ -1,15 +1,37 @@
-import type { RpcUrlsMap } from '@metamask/connect-multichain';
 import { getInfuraRpcUrls as getInfuraRpcUrlsMultichain } from '@metamask/connect-multichain';
 import type { CaipChainId, Hex } from '@metamask/utils';
 import {
+  hexToNumber,
   KnownCaipNamespace,
   numberToHex,
   parseCaipChainId,
 } from '@metamask/utils';
 
-export const getInfuraRpcUrls = (infuraAPIKey: string): RpcUrlsMap => {
-  const caipMap = getInfuraRpcUrlsMultichain(infuraAPIKey);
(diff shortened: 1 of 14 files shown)


packages/connect-evm/src/utils/infura.ts
```diff
@@ -1,15 +1,37 @@
...
+ * @param options.chainIds - Optional hex chain IDs to filter the output
+ * @returns A map of hex chain IDs to Infura RPC URLs
+ */
+export const getInfuraRpcUrls = ({
+  infuraApiKey,
+  chainIds,
+}: {
+  infuraApiKey: string;
+  chainIds?: Hex[];
+}): Record<Hex, string> => {
```
ffmcgee725, line 24: What do you think of changing the signature to have 2 params instead of the object param

```ts
export const getInfuraRpcUrls = (infuraApiKey: string,
  chainIds?: Hex[],
): Record<Hex, string> =>
```

That way we can make second argument optional and it wouldn't be a breaking change ?

jiexi: Great point. Alex D and I chatted about this more, and if it's just the breaking change you are concerned about then actually now is a great time to change this interface before we have actual consumers. We're going to merge as is for now. We still have time to adjust on Monday if you feel strongly about this. Thank you, sir!

adonesky1 (approved):

PR #230, opened 2026-03-17
refactor: unify connect notification flow and establish typed sdk event baseline

ffmcgee725:
## Explanation

- Remove the notification method filter in `DefaultTransport` so all wallet notifications are forwarded (matching MWP transport behavior)
- Add explicit `SDKEvents` entries for `metamask_accountsChanged`, `metamask_chainChanged`, and `stateChanged`
- Route `stateChanged` through `emit()` so `client.on('stateChanged', cb)` works correctly
- Migrate `ConnectEvm` from `transport.onNotification()` to typed `core.on()`/`core.off()`, removing all `@ts-expect-error` suppressions

## Motivation

The SDK had two parallel notification delivery paths — `core.on(...)` and `transport.onNotification()` — for what should be a single concern. This caused:

1. **Asymmetric transport behavior** — `DefaultTransport` silently dropped any notification that wasn't `metamask_chainChanged` or `metamask_accountsChanged`, while MWP forwarded everything.
2. **Ecosystem clients coupled to transport internals** — `ConnectEvm` subscribed directly to `transport.onNotification()` for chain/account events (with `@ts-expect-error` on every access) instead of using the core event bus.
3. **`stateChanged` silently failing** — The status setter emitted only via `options.transport?.onNotification?.(...)`, never via `super.emit()`, so `client.on('stateChanged', cb)` never fired.
4. **Unmodeled notification types** — `SDKEvents` had no explicit entries for key notification events; the catch-all index signature hid this gap.

### `packages/connect-multichain/src/multichain/transports/default/index.ts`
Removed the `metamask_chainChanged || metamask_accountsChanged` guard in `#handleNotification`. The transport now forwards any notification with a `method` property, with a generalized null/type guard.

### `packages/connect-multichain/src/domain/events/types/index.ts`
Added typed entries for `metamask_accountsChanged` (`string[]`), `metamask_chainChanged` (`{ chainId: string }`), and `stateChanged` (connection status union). The `ConnectionStatus` type is inlined to avoid a circular import. The catch-all index signature is retained for backward compatibility.

### `packages/connect-multichain/src/multichain/index.ts`
Changed the `status` setter to call `this.emit('stateChanged', value)` instead of `this.options.transport?.onNotification?.(...)`. The existing `emit` override already calls both `onNotification` and `super.emit`, so both paths are covered without duplication.

### `packages/connect-evm/src/connect.ts`
Replaced `this.#core.transport.onNotification(...)` with typed `this.#core.on('metamask_accountsChanged', ...)` and `this.#core.on('metamask_chainChanged', ...)`. Cleanup uses `this.#core.off(...)`. All four `@ts-expect-error` suppressions are removed.

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-1141

## Checklist

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Migrate `metamask_accountsChanged` and `metamask_chainChanged` subscriptions from `transport.onNotification()` to typed `core.on()`/`core.off()`, removing `@ts-expect-error` suppressions ([#230](https://github.com/MetaMask/connect-monorepo/pull/230))
+
 ## [0.8.0]
 
 ### Changed
--- a/packages/connect-evm/src/connect.test.ts
+++ b/packages/connect-evm/src/connect.test.ts
@@ -70,6 +70,11 @@ function createMockCore(): MockCore {
       }
       handlers[event].push(handler);
     },
+    off(event: string, handler: (...args: unknown[]) => void): void {
+      if (handlers[event]) {
+        handlers[event] = handlers[event].filter((fn) => fn !== handler);
+      }
+    },
     emit(event: string, ...args: unknown[]): void {
       handlers[event]?.forEach((handler) => handler(...args));
     },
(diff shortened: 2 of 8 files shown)


wenfix (approved):
Tested and looks good! 🔥

PR #234, opened 2026-03-18
fix: mark @react-native-async-storage/async-storage peer dep as optional

adonesky1:
## Summary

- Adds `peerDependenciesMeta` to `@metamask/connect-multichain` marking `@react-native-async-storage/async-storage` as `optional: true`
- The async-storage peer dependency is only used by the React Native build entry (`index.native.ts` → `store/adapters/rn.ts`). Browser and Node entry points (`index.browser.ts`, `index.node.ts`) never import it.
- Without the optional flag, package managers like pnpm eagerly resolve this peer dep, pulling in the **entire** `react-native` CLI/dev toolchain (`@react-native/debugger-frontend`, `chrome-launcher`, `chromium-edge-launcher`, `metro`, etc.) for web consumers
- This triggers 20+ false-positive security alerts from Socket on downstream PRs (e.g. [metamask-connect-examples#2](https://github.com/MetaMask/metamask-connect-examples/pull/2))

### Dependency chain causing alerts

```
@metamask/connect-evm
  → @metamask/connect-multichain
    ← @react-native-async-storage/async-storage (PEER DEP, not optional)
      → react-native@0.84.1
        → @react-native/community-cli-plugin
          → @react-native/dev-middleware
            → @react-native/debugger-frontend  ← HIGH: "obfuscated code"
            → chrome-launcher                  ← shell + network alerts
            → chromium-edge-launcher           ← shell + network alerts
```

### Why this is safe

1. The browser build entry (`src/index.browser.ts`) dynamically imports `store/adapters/web.ts` — never `rn.ts`
2. The node build entry (`src/index.node.ts`) dynamically imports `store/adapters/node.ts` — never `rn.ts`
3. Only `src/index.native.ts` imports the RN adapter
4. The tsup config already marks `@react-native-async-storage/async-storage` as external in all build targets
5. React Native consumers who actually need it will already have it installed as part of their RN project

## Test plan

- [x] Verify browser/node builds are unaffected (no import of async-storage)
- [ ] Verify React Native build still resolves async-storage when it's installed
- [ ] Regenerate lockfile in a web consumer (e.g. examples repo) and confirm react-native subtree is gone

### DIFF
--- a/packages/connect-multichain/src/index.native.ts
+++ b/packages/connect-multichain/src/index.native.ts
@@ -20,7 +20,15 @@ export const createMultichainClient: CreateMultichainFN = async (options) => {
   if (options.storage) {
     storage = options.storage;
   } else {
-    const { StoreAdapterRN } = await import('./store/adapters/rn');
+    let StoreAdapterRN;
+    try {
+      ({ StoreAdapterRN } = await import('./store/adapters/rn'));
+    } catch {
+      throw new Error(
+        '@metamask/connect-multichain: @react-native-async-storage/async-storage is required for React Native. ' +
+          'Install it with: npm install @react-native-async-storage/async-storage',
+      );
+    }
     const adapter = new StoreAdapterRN();
     storage = new Store(adapter);
   }
--- a/packages/connect-multichain/src/index.native.ts
+++ b/packages/connect-multichain/src/index.native.ts
@@ -26,8 +26,8 @@ export const createMultichainClient: CreateMultichainFN = async (options) => {
     } catch (error) {
       throw new Error(
         '@metamask/connect-multichain: @react-native-async-storage/async-storage is required for React Native. ' +
-          'Install it with: npm install @react-native-async-storage/async-storage',
-        { cause: error },
+          'Install it with: npm install @react-native-async-storage/async-storage\n' +
+          `Original error: ${error instanceof Error ? error.message : error}`,
       );
     }
     const adapter = new StoreAdapterRN();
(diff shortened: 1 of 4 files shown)


packages/connect-multichain/src/index.native.ts
```diff
@@ -20,7 +20,15 @@ export const createMultichainClient: CreateMultichainFN = async (options) => {
...
-    const { StoreAdapterRN } = await import('./store/adapters/rn');
+    let StoreAdapterRN;
+    try {
+      ({ StoreAdapterRN } = await import('./store/adapters/rn'));
+    } catch {
+      throw new Error(
+        '@metamask/connect-multichain: @react-native-async-storage/async-storage is required for React Native. ' +
+          'Install it with: npm install @react-native-async-storage/async-storage',
+      );
+    }
```
jiexi, line 32: is the native `Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'XYZ` error sufficient? Or should we at least be checking for that error specifically here? I'm afraid we might create additional confusion by not at least including the original error

adonesky1: Agreed — the original error is now preserved via `{ cause: error }` on the thrown `Error`. This way the actionable install hint is still front-and-center, but the underlying `ERR_MODULE_NOT_FOUND` (or whatever actually went wrong) is accessible on `.cause` for debugging.

jiexi (dismissed):

jiexi (approved):

PR #238, opened 2026-03-20
fix(platform): resolve ReactNative platform type for Hermes-based RN apps

adonesky1:
## Problem

https://consensyssoftware.atlassian.net/browse/WAPI-1259

In modern React Native apps using the **Hermes engine** (the default JS engine since RN 0.70), `window` is undefined at runtime. This caused `isReactNative()` in `packages/connect-multichain/src/domain/platform/index.ts` to silently return `false`, because it only checked `window.navigator.product === 'ReactNative'`:

```typescript
// Before — misses Hermes RN entirely
function isReactNative(): boolean {
  const hasWindowNavigator =
    typeof window !== 'undefined' && window.navigator !== undefined;
  const nav = hasWindowNavigator ? window.navigator : undefined;

  if (!nav) {
    return false; // ← Hermes RN exits here: window is undefined
  }
  return hasWindowNavigator && window.navigator?.product === 'ReactNative';
}
```

`isNotBrowser()` _did_ have the correct `global.navigator.product` check, but returned `true` (meaning "not a browser"), causing `getPlatformType()` to resolve to `PlatformType.NonBrowser` instead of `PlatformType.ReactNative`.

> **Note:** `navigator.product === 'ReactNative'` is the [established community pattern](https://stackoverflow.com/questions/39468022/how-do-i-know-if-my-code-is-running-as-react-native) for detecting a React Native environment in shared library code (as opposed to using React Native's own `Platform` API, which is not available in non-RN builds). This same detection failure has historically caused bugs in other libraries — the most widely-cited example being [axios/axios#2902](https://github.com/axios/axios/issues/2902), where `navigator.product` returning `undefined` in non-debug RN builds caused axios to incorrectly treat React Native as a standard browser environment.

### Impact: analytics silently disabled for React Native consumers

`#setupAnalytics()` in `connect-multichain` only calls `analytics.enable()` when the platform is a browser or `ReactNative`:

```typescript
if (!isBrowser && !isReactNative) {
  return; // analytics.enable() is never called
}
```

Because all modern (Hermes-based) React Native apps fell through to `NonBrowser`, analytics were **silently disabled** for all RN consumers despite React Native being an explicitly supported first-class platform (the event schema includes `'react-native'` as a valid `platform` value throughout).

Discovered while investigating [WAPI-1259](https://consensyssoftware.atlassian.net/browse/WAPI-1259).

### 1. `packages/connect-multichain/src/domain/platform/index.ts`

Move the `global.navigator.product` check from `isNotBrowser()` into `isReactNative()`, where it belongs. Since `getPlatformType()` calls `isReactNative()` first, Hermes RN environments now correctly resolve to `PlatformType.ReactNative`:

```typescript
function isReactNative(): boolean {
  // Modern Hermes-based RN: window is undefined, but global.navigator.product is set.
  // This check must come first so getPlatformType() resolves to ReactNative rather
  // than NonBrowser when isNotBrowser() would otherwise catch this case.
  if (
    typeof global !== 'undefined' &&
    global?.navigator?.product === 'ReactNative'
  ) {
    return true;
  }

  // Legacy RN environments where window === global
  const hasWindowNavigator =
    typeof window !== 'undefined' && window.navigator !== undefined;
  return hasWindowNavigator && window.navigator?.product === 'ReactNative';
}
```

### 2. `playground/react-native-playground/polyfills.ts`

The playground polyfill creates a bare `window = {}` without a `navigator` property. This was already handled by the `global.navigator` check above, but `window.navigator` being absent would still cause any code that reads `window.navigator` directly (outside of our fixed path) to miss the React Native context. Added the missing navigator polyfill for consistency and defence in depth:

```typescript
if (!windowObj.navigator) {
  windowObj.navigator = { product: 'ReactNative' };
}
```

### 3. `playground/react-native-playground/src/sdk/SDKProvider.tsx`

Added a startup log to make the platform detection result directly observable when running the playground:

```typescript
console.log('[Analytics] platform =', getPlatformType());
```

### Unit tests (automated)

Two test cases cover the library fix:

1. **Existing test** (`'should handle global navigator.product being ReactNative'`) — corrected expected value from `NonBrowser` to `ReactNative`.
2. **New Hermes-specific test** (`'should return ReactNative when window is undefined but global.navigator.product is ReactNative (Hermes engine)'`) — explicitly stubs `window` to `undefined` while setting `global.navigator.product = 'ReactNative'`, directly simulating the Hermes runtime scenario that was previously broken.

All 306 tests pass.

### E2E — platform detection (quick, ~5 min)

Run the RN playground on a physical device or simulator:

```bash
cd playground/react-native-playground
yarn ios    # or yarn android
```

Open Metro / the device log and filter for `[Analytics]`. You should see:

```
[Analytics] platform = react-native   ✅ (fix working)
```

A result of `nodejs` means the fix is not in effect.

### E2E — full analytics pipeline (thorough, ~15 min)

To verify that events actually reach the backend after the fix:

1. Run the playground on a device with [Flipper](https://fbflipper.com/) (requires a dev build, not Expo Go) and open the **Network** plugin, or configure [Charles Proxy](https://www.charlesproxy.com/) on the device.
2. Filter traffic for `mm-sdk-analytics.api.cx.metamask.io`.
3. Trigger any action in the playground that would fire an analytics event (e.g. initiate a wallet connection).
4. Confirm a `POST /v2/events` request appears with `"platform": "react-native"` in the body.

Before the fix: no requests. After the fix: requests appear.

## References

- [SO: How do I know if my code is running as React Native](https://stackoverflow.com/questions/39468022/how-do-i-know-if-my-code-is-running-as-react-native) — the `navigator.product` detection pattern, community consensus
- [axios/axios#2902](https://github.com/axios/axios/issues/2902) — real-world precedent: `navigator.product` returning `undefined` in non-debug RN builds breaks platform detection in shared libraries
- [facebook/react-native#1331](https://github.com/facebook/react-native/issues/1331) — React Native's own discussion of how to detect the RN runtime without importing the `Platform` module
- [React Native: Using Hermes](https://reactnative.dev/docs/next/hermes) — `global.HermesInternal` as the canonical Hermes engine detection check; Hermes is the default engine since RN 0.70
- [WAPI-1259](https://consensyssoftware.atlassian.net/browse/WAPI-1259) — the investigation that surfaced this issue


ffmcgee725 (dismissed):

packages/connect-multichain/src/domain/index.test.ts
```diff
@@ -233,8 +233,31 @@ t.describe('Platform Detection', () => {
...
+        // Before the fix, this resolved to NonBrowser, disabling analytics for all RN consumers.
+        t.vi.stubGlobal('window', undefined);
+        t.vi.stubGlobal('global', {
+          navigator: {
+            product: 'ReactNative',
+          },
+        });
+        t.vi.stubGlobal('navigator', {
+          product: 'ReactNative',
+        });
```
jiexi: is this navigator stub needed?

adonesky1: Removed the redundant navigator stub here and in the earlier global navigator.product test so both cases now exercise the intended global.navigator.product === "ReactNative" path directly.

Commit: https://github.com/MetaMask/connect-monorepo/pull/238/commits/371826d7099414e31e16a30c1f133ce41e2e622f

playground/react-native-playground/polyfills.ts
```diff
@@ -27,6 +27,14 @@ if (typeof global !== 'undefined' && global.window) {
   windowObj = {};
 }
 
+// Ensure navigator exists with the React Native product identifier.
+// Without this, isReactNative() in connect-multichain falls back to checking
+// global.navigator (which works), but window.navigator.product would be missing,
+// meaning any code that reads window.navigator directly would not detect RN correctly.
+if (!windowObj.navigator) {
+  windowObj.navigator = { product: 'ReactNative' };
+}
```
jiexi: I don't quite understand why we need a polyfill. Shouldn't this property exist already since this is running in the react native environment?

adonesky1: The `window` object itself doesn't exist in React Native — the polyfill creates it from scratch. It's needed because `connect-multichain` uses `window` unconditionally at module load time: EIP-6963 provider discovery fires `window.addEventListener` and `window.dispatchEvent` in an IIFE the moment the module is imported, and `multichain/index.ts` attaches a `window.addEventListener('beforeunload')` listener for cleanup. Without `window`, the SDK throws on startup in RN before any user code runs.

So the polyfill creates a fake `window = {}` with just enough surface area (`addEventListener`, `dispatchEvent`, `location`) to keep those paths from throwing. The `navigator` property just happened to be missing from that fake object — this PR adds it so `window.navigator` and `global.navigator` stay consistent.

adonesky1: `window` doesn't exist natively in RN, so this polyfill creates it. It's required because `connect-multichain` calls `window.addEventListener` and `window.dispatchEvent` unconditionally at module load time (EIP-6963 discovery runs in an IIFE on import) — without it the SDK throws before any user code runs.

The fake `window = {}` just happened to be missing `navigator`, so `window.navigator` and `global.navigator` were inconsistent. This PR fills that gap.

jiexi (approved):

PR #239, opened 2026-03-20
chore: turborepo

wenfix:
## Summary

Replaces `yarn workspaces foreach` with [Turborepo](https://turbo.build/repo) for build and test orchestration across the monorepo. Turborepo understands the workspace dependency graph and adds local task caching — unchanged packages are skipped entirely on repeat runs.

This is a pure infrastructure change. No package APIs, exports, or behaviour change.

## What changed

- **`turbo.json`** — new root config defining the task pipeline and per-package cache output declarations
- **Root `package.json`** — `build`, `test:verbose`, `test:packages`, `build:docs`, `changelog:*` scripts now delegate to `turbo run`
- **`packages/connect-evm` and `packages/connect-multichain`** — removed `pretest` scripts that manually built upstream dependencies before tests; Turbo's `dependsOn: ["^build"]` handles this automatically and correctly
- **`.github/workflows/build-lint-test.yml`** — simplified CI:
  - deleted the `prepare` job that enumerated workspace names for the matrix
  - collapsed `validate-changelog` from a per-package matrix to a single `turbo run` call
  - reduced the `test` matrix from `N packages × 3 Node versions` to `3 jobs` (one per Node version), each running `turbo run test` across all packages
  - added `actions/cache@v4` on `.turbo` for build, test, and changelog jobs
- **`publish-release.yml` / `publish-preview.yml`** — added `.turbo` cache step before `yarn build`
- **`docs/contributing.md`** — updated "Performing operations across the monorepo" section to document Turbo; `yarn turbo run <task> --filter=<pkg>` is now the recommended way to run a task for a single package (uses the pinned local binary, gets caching, builds dependencies first)

## Why Turborepo over Nx

- Drop-in: one config file, no per-package changes, works natively with Yarn 4 workspaces
- No infra required: remote caching is deferred; local + GitHub Actions cache is sufficient today
- Right-sized: Nx adds generators, executors, and plugins — useful at a different scale, overkill for 6 library packages

## Performance

| Scenario                     | Before                  | After                                    |
| ---------------------------- | ----------------------- | ---------------------------------------- |
| Full cold build              | ~26s                    | ~20s                                     |
| Repeat build (no changes)    | ~26s                    | **~1.2s** (17× faster)                   |
| CI test jobs                 | `N packages × 3` matrix | `3 jobs` flat                            |
| PR touching only `analytics` | rebuilds everything     | rebuilds analytics + its dependents only |

Caching is filesystem-local per runner today. Cross-branch cache hits come from the `restore-keys` prefix match on `actions/cache`. Remote caching (Vercel or self-hosted) can be layered on with no further code changes if we want full cross-PR cache sharing later.

## No changes to

- Individual package APIs or exports
- TypeScript configs
- Yarn 4 workspace config
- `lavamoat` / `simple-git-hooks` / `depcheck` setup
- Publish / release workflows (logic unchanged; only a cache step added)
- Deploy-to-gh-pages workflows (separate follow-up)

## Checklist


wenfix:
One of the coolest aspects is compacting up CI jobs: turbo will run the test command for every package that supports it in the same workflow, in a non-blocking way and respecting dependency graphs.

Screenshot 2026-03-20 at 10 26 48

.github/workflows/build-lint-test.yml
```diff
@@ -110,19 +87,24 @@ jobs:
...
-    needs: prepare
     strategy:
       matrix:
         node-version: [20.x, 22.x, 24.x]
-        package-name: ${{ fromJson(needs.prepare.outputs.child-workspace-package-names) }}
     steps:
       - name: Checkout and setup environment
         uses: MetaMask/action-checkout-and-setup@v1
         with:
           is-high-risk-environment: false
```
adonesky1, line 99: The matrix + job title imply we're testing 20/22/24, but `node-version` never gets passed into checkout — so all three jobs are almost certainly the same Node from `.nvmrc`. That also means three separate turbo cache keys for the same work.

Can we add `node-version: ${{ matrix.node-version }}` under `with:` here?

turbo.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "lint": {
+      "outputs": []
+    },
+    "build:docs": {
+      "dependsOn": ["build"],
+      "outputs": ["docs/**"]
+    },
+    "changelog:validate": {
+      "outputs": []
```
adonesky1, line 35: Nit: without `inputs`, turbo will treat *any* file change in the package as a changelog-task invalidation. Scoping to `CHANGELOG.md` (and maybe `package.json`) would make cache hits way more likely for changelog-only CI.

turbo.json
```diff
@@ -0,0 +1,50 @@
...
+    "test:verbose": {
+      "dependsOn": ["^build"],
+      "outputs": []
+    },
+    "test:ci": {
+      "dependsOn": ["^build"],
+      "outputs": []
+    },
+    "lint": {
+      "outputs": []
```
adonesky1: `lint` is defined here but nothing actually runs `turbo run lint` — root `yarn lint` is its own thing. Might confuse the next person; either drop it or leave a one-liner in the PR desc / comment in file.

turbo.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "changelog:format": {
+      "outputs": []
+    },
+    "changelog:update": {
+      "outputs": []
+    },
+    "clean": {
+      "cache": false
```
adonesky1: Minor: `clean` doesn't look wired up anywhere (`yarn clean` is still the big rimraf). Could delete this block or hook it up later — just don't want folks thinking `turbo run clean` does something today.

packages/analytics/package.json
```diff
@@ -45,7 +45,7 @@
     "prepare-manifest:preview": "../../scripts/prepare-preview-manifest.sh",
     "publish:preview": "yarn npm publish --tag preview",
     "since-latest-release": "../../scripts/since-latest-release.sh",
-    "test": "vitest",
+    "test": "vitest run",
```
adonesky1, line 48: `vitest run` here is the right call under turbo 👍

Random consistency nit while we're in this file: `build` still does `yarn clean &&` but connect* dropped that for turbo — not urgent, just uneven if we want one pattern.

package.json
```diff
@@ -34,9 +34,9 @@
     "prepare-release": "yarn create-release-branch -i && yarn bump-playground-versions",
     "publish-previews": "yarn workspaces foreach --all --no-private --parallel --verbose run publish:preview",
     "test": "yarn test:scripts --silent && yarn test:packages",
-    "test:clean": "yarn workspaces foreach --all --parallel --verbose run test:clean && yarn test",
-    "test:packages": "yarn test:verbose --silent",
-    "test:verbose": "yarn workspaces foreach --all --parallel --verbose run test:verbose",
+    "test:clean": "yarn build:only-clean && turbo run test",
+    "test:packages": "turbo run test:verbose",
```
adonesky1, line 38: `test:packages` vs CI `turbo run test` aside — `publish-previews` is still the lone `workspaces foreach`, which is fine (publish isn't really a turbo task). A short note in contributing explaining why that one stayed on foreach would preempt future "why not turbo everything" threads.

.github/workflows/build-lint-test.yml
```diff
@@ -48,17 +24,19 @@ jobs:
...
-      - run: yarn workspace ${{ matrix.package-name }} changelog:validate
+      - name: Restore Turbo cache
+        uses: actions/cache@v4
+        with:
+          path: .turbo
+          key: turbo-${{ runner.os }}-changelog-${{ github.sha }}
+          restore-keys: |
+            turbo-${{ runner.os }}-changelog-
+      - run: yarn changelog:validate
```
adonesky1, line 39: Turbo will hit every workspace with `changelog:validate`, including private ones — old flow skipped those via `--no-private`. Looks intentional (node-playground changelog tweak backs that up). 👍

Ultra-nit: restoring `.turbo` here buys almost nothing for a no-output task; harmless, just mostly noise vs build/test.

packages/connect-multichain/package.json
```diff
@@ -51,7 +51,7 @@
   ],
   "scripts": {
     "allow-scripts": "",
-    "build": "yarn clean && tsc -b tsconfig.build.json && npx tsup",
+    "build": "tsc -b tsconfig.build.json && npx tsup",
```
adonesky1, line 54: Dropping `yarn clean` is the right tradeoff for turbo-cached builds. Only caveat: if someone runs `yarn workspace … build` *outside* turbo, stale `dist/` junk could linger — `yarn build:clean` / docs mention is enough for most of us.

packages/connect/package.json
```diff
@@ -56,7 +56,7 @@
     "dist/"
   ],
   "scripts": {
-    "build": "yarn clean && tsc -b --force tsconfig.build.json && npx tsup src/index.ts src/multichain/index.ts src/evm/index.ts --format esm,cjs --external @metamask/connect-multichain --external @metamask/connect-evm",
+    "build": "tsc -b --force tsconfig.build.json && npx tsup src/index.ts src/multichain/index.ts src/evm/index.ts --format esm,cjs --external @metamask/connect-multichain --external @metamask/connect-evm",
```
adonesky1, line 59: Same thought as connect-multichain — without `yarn clean`, direct workspace builds aren't a full reset anymore; turbo + `build:clean` is the happy path.

.github/workflows/build-lint-test.yml
wenfix: Fixed — added `node-version: ${{ matrix.node-version }}` to the checkout step. Pre-existing gap carried over from the original workflow.

turbo.json
wenfix: Fixed — added `inputs: ["CHANGELOG.md", "package.json"]` to `changelog:validate`.

turbo.json
wenfix: Dropped — no package in the monorepo has a `lint` script so this was dead config.

turbo.json
wenfix: Dropped.

packages/analytics/package.json
wenfix: Fixed — removed `yarn clean &&` from the build script. This was missed in the earlier fix commit that only updated the `test` script.

package.json
wenfix: Added a note in `docs/contributing.md` explaining why `publish-previews` stays on `yarn workspaces foreach`. On the `test:packages` vs CI `turbo run test` asymmetry — intentional: `test:verbose` for local dev output, `test` for CI pass/fail.

.github/workflows/build-lint-test.yml
wenfix: Kept — now more useful with the `inputs` scoping added to `changelog:validate`, which means cache hits are realistic on unchanged changelogs.

packages/connect-multichain/package.json
wenfix: Noted — `yarn build:clean` is the escape hatch for a full reset, and the contributing guide already documents it as the recommended clean build path.

packages/connect/package.json
wenfix: Same answer as connect-multichain.

.github/workflows/publish-preview.yml
```diff
@@ -43,6 +43,13 @@
         id: commit-sha
         run: echo "COMMIT_SHA=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
       - run: yarn prepare-preview-builds @metamask-previews ${{ steps.commit-sha.outputs.COMMIT_SHA }}
+      - name: Restore Turbo cache
+        uses: actions/cache@v4
```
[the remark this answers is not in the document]
wenfix: Removed — cache steps dropped from both publish workflows. The build speedup isn't worth the attack surface on privileged workflows.

.github/workflows/publish-release.yml
```diff
@@ -25,6 +25,13 @@
       - uses: MetaMask/action-publish-release@v3
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Restore Turbo cache
+        uses: actions/cache@v4
```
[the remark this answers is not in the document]
wenfix: Removed — cache steps dropped from both publish workflows. The build speedup isn't worth the attack surface on privileged workflows.

playground/node-playground/CHANGELOG.md
```diff
@@ -7,16 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
-
-- Update `getInfuraRpcUrls` call to use new options object parameter ([#211](https://github.com/MetaMask/connect-monorepo/pull/211))
-
```
jiexi, line 13: my bad. thank you!

adonesky1 (approved):
LGTM!

PR #243, opened 2026-03-26
fix: allow MWP to accept subsequent connect requests

wenfix:
## Explanation

When a user wants to add or change accounts after an initial connection, consumers like Portfolio currently have to `disconnectAsync()` + `connectAsync()` through wagmi. This forces the user through a full reconnection flow — bad UX, especially on mobile where it means scanning a QR code again.

The `wallet_requestPermissions` flow in `connect-evm` is designed to handle this by calling `connect({ forceRequest: true })`, which should bypass the "already connected" short-circuit and re-prompt the user for account selection. The `forceRequest` plumbing already exists at every layer — `connect-evm`, `MetaMaskConnectMultichain`, `DefaultTransport` — except the final one: **`MWPTransport` ignores `forceRequest`**, so the re-prompt silently no-ops on mobile.

### Changes

**1. `MWPTransport.connect()` — accept `forceRequest`** (`connect-multichain`)

Added `forceRequest?: boolean` to the `connect()` options type so it's forwarded to `onResumeSuccess`.

**2. `MWPTransport.onResumeSuccess()` — bypass scope check when forced** (`connect-multichain`)

When `forceRequest` is true, skip the `isSameScopesAndAccounts` check and unconditionally send `wallet_createSession`. This mirrors `DefaultTransport`'s existing behavior.

**3. Test: `forceRequest` is forwarded on `wallet_requestPermissions`** (`connect-evm`)

Added a test verifying that `wallet_requestPermissions` calls `core.connect` with `forceRequest: true`.

### Flow after fix

```
wallet_requestPermissions
  → connect-evm interceptor: connect({ chainIds: [DEFAULT_CHAIN_ID], forceRequest: true })
  → MetaMaskConnectMultichain.connect() forwards forceRequest to transport
  → MWPTransport.onResumeSuccess() skips scope check, sends wallet_createSession
  → MetaMask Mobile presents account selection UI
  → wallet responds with updated session
  → wallet_sessionChanged fires
  → accountsChanged emitted → wagmi state updates
```

## References

Needed to fix [WAPI-1349](https://consensyssoftware.atlassian.net/browse/WAPI-1349)

## Checklist

- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/connect-monorepo/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `wallet_requestPermissions` now correctly re-prompts account selection on MWP (mobile) without requiring a full disconnect/reconnect cycle. The request forwards `forceRequest: true` through to `MWPTransport`, which previously ignored it and silently returned the existing session. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
+- `wallet_requestPermissions` now uses the current session's chain IDs when re-prompting, instead of hardcoding mainnet (`0x1`). This preserves scope parity with the original connection. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
+- `#onSessionChanged` now always derives accounts from session scopes via `getEthAccounts()` rather than making an `eth_accounts` RPC call when connected. On MWP, the `eth_accounts` cache may not yet reflect the updated session at the time `wallet_sessionChanged` fires, causing stale accounts to be emitted. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
+
 ## [0.9.0]
 
 ### Changed
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -10,8 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - `wallet_requestPermissions` now correctly re-prompts account selection on MWP (mobile) without requiring a full disconnect/reconnect cycle. The request forwards `forceRequest: true` through to `MWPTransport`, which previously ignored it and silently returned the existing session. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
-- `wallet_requestPermissions` now uses the current session's chain IDs when re-prompting, instead of hardcoding mainnet (`0x1`). This preserves scope parity with the original connection. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
-- `#onSessionChanged` now always derives accounts from session scopes via `getEthAccounts()` rather than making an `eth_accounts` RPC call when connected. On MWP, the `eth_accounts` cache may not yet reflect the updated session at the time `wallet_sessionChanged` fires, causing stale accounts to be emitted. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
 
 ## [0.9.0]
 
--- a/packages/connect-evm/src/connect.test.ts
+++ b/packages/connect-evm/src/connect.test.ts
@@ -195,15 +195,10 @@ describe('MetamaskConnectEVM', () => {
         );
       });
 
-      it('connects using accounts from a eth_accounts response when the MultichainClient is connected', async () => {
+      it('connects using accounts from session scopes when the MultichainClient is connected', async () => {
         const mockCore = createMockCore();
         mockCore._status = 'connected';
         mockCore.storage.adapter.get.mockResolvedValue(JSON.stringify('0x1'));
-        mockCore.transport.sendEip1193Message.mockResolvedValue({
-          result: ['0xabcdefabcdefabcdefabcdefabcdefabcdefabcd'],
-          id: 1,
-          jsonrpc: '2.0',
-        });
 
         const client = await MetamaskConnectEVM.create({ core: mockCore });
 
@@ -228,9 +223,9 @@ describe('MetamaskConnectEVM', () => {
         const connectData = await connectPromise;
         expect(connectData.chainId).toBe('0x1');
         expect(connectData.accounts).toContain(
-          '0xabcdefabcdefabcdefabcdefabcdefabcdefabcd',
+          '0x1234567890123456789012345678901234567890',
         );
-        expect(mockCore.transport.sendEip1193Message).toHaveBeenCalledWith({
+        expect(mockCore.transport.sendEip1193Message).not.toHaveBeenCalledWith({
           method: 'eth_accounts',
           params: [],
         });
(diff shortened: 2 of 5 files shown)


ffmcgee725 (dismissed):

adonesky1:
Tested on the test dapp and validated that `wallet_requestPermissions` was not working for already connected MWP sessions before, but is now 🚀

packages/connect-evm/src/connect.test.ts
```diff
@@ -228,9 +223,9 @@ describe('MetamaskConnectEVM', () => {
         const connectData = await connectPromise;
         expect(connectData.chainId).toBe('0x1');
         expect(connectData.accounts).toContain(
-          '0xabcdefabcdefabcdefabcdefabcdefabcdefabcd',
+          '0x1234567890123456789012345678901234567890',
         );
-        expect(mockCore.transport.sendEip1193Message).toHaveBeenCalledWith({
+        expect(mockCore.transport.sendEip1193Message).not.toHaveBeenCalledWith({
```
adonesky1: should we keep this assertion at all? If we do we should probably make it more general that `sendEip1193Message` is not called at all rather than "with" some particular params

packages/connect-evm/src/connect.ts
```diff
@@ -757,17 +759,7 @@ export class MetamaskConnectEVM {
...
-      if (this.#core.status === 'connected') {
-        const ethAccountsResponse =
-          await this.#core.transport.sendEip1193Message({
-            method: 'eth_accounts',
-            params: [],
-          });
-        initialAccounts = ethAccountsResponse.result as Address[];
-      } else {
-        initialAccounts = getEthAccounts(this.#sessionScopes);
-      }
```
adonesky1, line 770: 🤔

adonesky1: I'm trying to remember why we needed the `if (this.#core.status === 'connected') {` block in the first place

adonesky1: the `#onTransportNotification` [logic](https://github.com/MetaMask/connect-monorepo/blob/99b2209c8a9e6309eefbfe4d5dd378b8b3a5f51d/packages/connect-multichain/src/multichain/index.ts#L262-L282) seems to imply that this `this.#core.status` will always either be `connected` or `disconnected` when it hits here

adonesky1: And if the status is 'disconnected' it would always be [an empty session object](https://github.com/MetaMask/connect-monorepo/blob/99b2209c8a9e6309eefbfe4d5dd378b8b3a5f51d/packages/connect-multichain/src/multichain/index.ts#L1047-L1048)

adonesky1: this branch was added in @jiexi 's singleton PR https://github.com/MetaMask/connect-monorepo/pull/157

adonesky1: oh was this about account ordering? I guess perhaps @jiexi 's changes to that front will actually make this a safe change... but we need to validate

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -427,6 +431,7 @@ export class MWPTransport implements ExtendedTransport {
     scopes: Scope[];
     caipAccountIds: CaipAccountId[];
     sessionProperties?: SessionProperties;
+    forceRequest?: boolean;
```
adonesky1, line 434: Lets add a test for this?

packages/connect-evm/src/connect.ts
```diff
@@ -617,14 +617,16 @@ export class MetamaskConnectEVM {
...
+      const chainIds: Hex[] =
+        sessionChainIds.length > 0 ? sessionChainIds : [DEFAULT_CHAIN_ID];
+      const scope: Scope = `eip155:${chainIds[0]}`;
 
       await this.#trackWalletActionRequested(method, scope, params);
 
       try {
         const result = await this.connect({
-          chainIds: [initiallySelectedChainId],
+          chainIds,
```
adonesky1: couldn't this result in the chainId unexpectedly switching on a re-prompt?

packages/connect-evm/src/connect.ts
jiexi: This change doesn't seem relevant to the goal of the PR. Coincidentally, this change is what we need to make later after extension and mobile. both have wallet_sessionChanged account ordering AND we have majority user adoption of the versions those ship in.

packages/connect-evm/src/connect.test.ts
wenfix: Reverted this change — the test now asserts `sendEip1193Message` **was** called (original behavior restored), so this is no longer applicable.

packages/connect-evm/src/connect.ts
wenfix: Agreed — reverted this change. The `#onSessionChanged` simplification and the session-chain-ID interceptor change are both out of scope for this PR.

packages/connect-evm/src/connect.ts
wenfix: Reverted — back to using `DEFAULT_CHAIN_ID`. This avoids the unexpected chain switch on re-prompt and keeps the PR focused on the `forceRequest` plumbing.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
wenfix: Added a test in `connect.test.ts` that verifies `wallet_requestPermissions` passes `forceRequest: true` as the 4th arg to `core.connect`. Testing `forceRequest` directly through `MWPTransport.connect()` would require substantial mocking of the dappClient/session/resume flow — the existing MWP tests only cover `handleMessage`. The connect-evm level test covers the integration path.

packages/connect-evm/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `wallet_requestPermissions` now correctly re-prompts account selection on MWP (mobile) without requiring a full disconnect/reconnect cycle. The request forwards `forceRequest: true` through to `MWPTransport`, which previously ignored it and silently returned the existing session. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
+- `wallet_requestPermissions` now uses the current session's chain IDs when re-prompting, instead of hardcoding mainnet (`0x1`). This preserves scope parity with the original connection. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
+- `#onSessionChanged` now always derives accounts from session scopes via `getEthAccounts()` rather than making an `eth_accounts` RPC call when connected. On MWP, the `eth_accounts` cache may not yet reflect the updated session at the time `wallet_sessionChanged` fires, causing stale accounts to be emitted. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
```
jiexi:
```suggestion
```

packages/connect-evm/src/connect.ts
adonesky1: Yeah, defaulting to [DEFAULT_CHAIN_ID] fixes “random first permitted chain” but it’s the same bug through a different door: connect() always treats chainIds[0] as #pendingPreferredChainId, and that wins over the cached chain if it’s still permitted — so you’ll snap back to mainnet on every wallet_requestPermissions even when the user was on Polygon.

Better middle ground: keep the full permitted list for scope parity, but put the current chain first when it’s still in the session. Use provider state (and fall back if needed):
```
const permitted =
  getPermittedEthChainIds(this.#sessionScopes).length > 0
    ? getPermittedEthChainIds(this.#sessionScopes)
    : [DEFAULT_CHAIN_ID];
const selected = this.#provider.selectedChainId;
const preferred =
  selected && permitted.includes(selected) ? selected : permitted[0];
const chainIds = [
  preferred,
  ...permitted.filter((id) => id !== preferred),
];
const scope: Scope = `eip155:${preferred}`; // match analytics to actual chain
await this.connect({ chainIds, forceRequest: shouldForceConnectionRequest });
```

That keeps this PR’s intent (don’t reorder-only via raw getPermittedEthChainIds order) without resetting the user’s chain on re-prompt.

adonesky1: PR pointed at this one to make this change here: https://github.com/MetaMask/connect-monorepo/pull/245/

adonesky1 (approved):
LGTM!

PR #244, opened 2026-03-27
refactor: Lazy-load MWP packages + eciesjs/KeyManager behind dynamic `import()`

ffmcgee725:
## Explanation

Converts static imports of `@metamask/mobile-wallet-protocol-core`, `@metamask/mobile-wallet-protocol-dapp-client`, and `eciesjs` to dynamic `import()` calls so the entire MWP + crypto dependency tree is only loaded when MWP transport (QR code / deeplink pairing with MetaMask Mobile) is actually used. Desktop browser extension users never hit this code path and will no longer pay the load-time cost.

### Transitive dependencies now deferred

| Package | Brings in |
|---|---|
| `mwp-core` | `centrifuge` → `protobufjs` (12+ sub-packages incl. `@types/node`), `async-mutex`, `events` |
| `eciesjs` | `@ecies/ciphers`, `@noble/ciphers` (3 unique packages; `@noble/curves` / `@noble/hashes` shared with mwp-core) |

### `packages/connect-multichain/src/multichain/index.ts`

- **Removed top-level value imports** of `ErrorCode`, `ProtocolError`, `SessionStore`, `WebSocketTransport` (mwp-core), `DappClient` (mwp-dapp-client), `MWPTransport` (local transport), and `keymanager` (KeyManager).
- **Kept / added type-only imports** for `SessionRequest` and `DappClient` (erased at compile time).
- **`#createDappClient()`** now dynamically imports all three MWP packages in parallel via `Promise.all`, then calls the new `createKeyManager()` async factory.
- **Added `#mwpModules` class field** — caches `ProtocolError` and `ErrorCode` from mwp-core after the first lazy load so `error instanceof ProtocolError` and `ErrorCode.REQUEST_EXPIRED` checks in catch blocks continue to work correctly.
- **Added `#transportType` class field** — replaces all `instanceof MWPTransport` checks with a tracked `TransportType` enum, eliminating the need for a static import of the `MWPTransport` class.

### `packages/connect-multichain/src/multichain/transports/mwp/KeyManager.ts`

- Converted from a class with a static `eciesjs` import + exported singleton to an **async factory function** (`createKeyManager()`) that dynamically imports `eciesjs` and returns an `IKeyManager` via closure. This prevents esbuild from hoisting the `eciesjs` external import to the bundle's top level.

### `packages/connect-multichain/src/multichain/transports/mwp/index.ts`

- Converted the static `import { SessionStore }` from mwp-core to a dynamic `import()` inside `getActiveSession()`. This was the last remaining static import that esbuild was hoisting to the top level.

## Bundle size analysis

| Format | Before | After | Delta |
|---|---|---|---|
| browser/es (ESM) | 138,433 B | 145,724 B | +7,291 B (+5.3%) |
| browser/umd | 141,811 B | 148,868 B | +7,057 B (+5.0%) |
| browser/iife | 1,592,709 B | 1,631,104 B | +38,395 B (+2.4%) |
| node/cjs | 139,590 B | 146,647 B | +7,057 B (+5.1%) |
| node/es | 136,008 B | 143,299 B | +7,291 B (+5.4%) |
| react-native/es | 134,320 B | 141,953 B | +7,633 B (+5.7%) |

The package's own bundle increases ~5–7 KB due to esbuild's async module init wrappers (`__esm` blocks). This is expected because MWP packages are currently bundled (not external) with `splitting: false`.

**The key improvement is for downstream consumers**: the ESM output now contains **zero** top-level static imports of mwp-core, mwp-dapp-client, or eciesjs. A consumer's bundler (webpack, Vite, etc.) can code-split the entire MWP + crypto tree into a lazy chunk that is never loaded for extension-only users.

> **Follow-up opportunity**: adding MWP packages to the `external` list in `tsup.config.ts` or enabling `splitting: true` for ESM builds would let the package's own build benefit from the split as well.

## Test plan

- [x] All 306 existing tests pass (4 pre-existing skips)
- [x] TypeScript build succeeds with full type checking
- [x] No linter errors introduced
- [x] `error instanceof ProtocolError` checks still work correctly via cached `#mwpModules`
- [x] Verified ESM build output has no top-level static imports of MWP packages or eciesjs

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-1270

## Checklist

### DIFF
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Lazy-load MWP transport dependencies: `@metamask/mobile-wallet-protocol-core`, `@metamask/mobile-wallet-protocol-dapp-client`, and `eciesjs` are now dynamically imported only when MWP transport is actually used, allowing bundlers to code-split the entire MWP + crypto dependency tree for consumers who only use the browser extension flow ([#244](https://github.com/MetaMask/connect-monorepo/pull/244))
+
 ### Fixed
 
 - Fix platform detection for Hermes-based React Native apps: `isReactNative()` now checks `global.navigator.product` before `window.navigator.product`, so modern RN environments where `window` is undefined correctly resolve to `PlatformType.ReactNative` instead of `PlatformType.NonBrowser`. This restores analytics for all React Native consumers — analytics were silently disabled because `#setupAnalytics()` only calls `analytics.enable()` for browser and `ReactNative` platforms ([#238](https://github.com/MetaMask/connect-monorepo/pull/238))
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Lazy-load MWP transport dependencies: `@metamask/mobile-wallet-protocol-core`, `@metamask/mobile-wallet-protocol-dapp-client`, and `eciesjs` are now dynamically imported only when MWP transport is actually used, allowing bundlers to code-split the entire MWP + crypto dependency tree for consumers who only use the browser extension flow ([#244](https://github.com/MetaMask/connect-monorepo/pull/244))
+
 ## [0.12.1]
 
 ### Fixed
@@ -18,10 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Include `analytics.remote_session_id` in V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events ([#256](https://github.com/MetaMask/connect-monorepo/pull/256))
-
-### Changed
-
-- Lazy-load MWP transport dependencies: `@metamask/mobile-wallet-protocol-core`, `@metamask/mobile-wallet-protocol-dapp-client`, and `eciesjs` are now dynamically imported only when MWP transport is actually used, allowing bundlers to code-split the entire MWP + crypto dependency tree for consumers who only use the browser extension flow ([#244](https://github.com/MetaMask/connect-monorepo/pull/244))
 
 ## [0.11.1]
 
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -15,17 +15,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Lazy-load MWP transport dependencies: `@metamask/mobile-wallet-protocol-core`, `@metamask/mobile-wallet-protocol-dapp-client`, and `eciesjs` are now dynamically imported only when MWP transport is actually used, allowing bundlers to code-split the entire MWP + crypto dependency tree for consumers who only use the browser extension flow ([#244](https://github.com/MetaMask/connect-monorepo/pull/244))
 
-### Fixed
-
-- `MWPTransport.connect()` now accepts and forwards a `forceRequest` option. When `true`, `onResumeSuccess` skips the `isSameScopesAndAccounts` check and unconditionally sends `wallet_createSession`, allowing consumers to re-prompt account selection on an existing MWP session. Previously `forceRequest` was silently ignored, causing `wallet_requestPermissions` to no-op on mobile. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
-- Fix platform detection for Hermes-based React Native apps: `isReactNative()` now checks `global.navigator.product` before `window.navigator.product`, so modern RN environments where `window` is undefined correctly resolve to `PlatformType.ReactNative` instead of `PlatformType.NonBrowser`. This restores analytics for all React Native consumers — analytics were silently disabled because `#setupAnalytics()` only calls `analytics.enable()` for browser and `ReactNative` platforms ([#238](https://github.com/MetaMask/connect-monorepo/pull/238))
-
 ## [0.11.1]
 
 ### Changed
 
 - chore: align sub-package licenses with root ConsenSys 2022 license ([#241](https://github.com/MetaMask/connect-monorepo/pull/241))
 - chore: turborepo ([#239](https://github.com/MetaMask/connect-monorepo/pull/239))
+
+### Fixed
+
+- `MWPTransport.connect()` now accepts and forwards a `forceRequest` option. When `true`, `onResumeSuccess` skips the `isSameScopesAndAccounts` check and unconditionally sends `wallet_createSession`, allowing consumers to re-prompt account selection on an existing MWP session. Previously `forceRequest` was silently ignored, causing `wallet_requestPermissions` to no-op on mobile. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
+- Fix platform detection for Hermes-based React Native apps: `isReactNative()` now checks `global.navigator.product` before `window.navigator.product`, so modern RN environments where `window` is undefined correctly resolve to `PlatformType.ReactNative` instead of `PlatformType.NonBrowser`. This restores analytics for all React Native consumers — analytics were silently disabled because `#setupAnalytics()` only calls `analytics.enable()` for browser and `ReactNative` platforms ([#238](https://github.com/MetaMask/connect-monorepo/pull/238))
 
 ## [0.11.0]
 
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -82,12 +82,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
 
   #beforeUnloadListener: (() => void) | undefined;
 
-  #mwpModules?: Pick<
-    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
-    typeof import('@metamask/mobile-wallet-protocol-core'),
-    'ProtocolError' | 'ErrorCode'
-  >;
-
   #transportType?: TransportType;
 
   public _status: ConnectionStatus = 'pending';
@@ -300,10 +294,8 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       } else if (transportType === TransportType.MWP) {
         const { adapter: kvstore } = this.options.storage;
         const dappClient = await this.#createDappClient();
-        const { MWPTransport: MWPTransportClass } = await import(
-          './transports/mwp'
-        );
-        const apiTransport = new MWPTransportClass(dappClient, kvstore);
+        const { MWPTransport } = await import('./transports/mwp');
+        const apiTransport = new MWPTransport(dappClient, kvstore);
         this.#dappClient = dappClient;
         this.#transport = apiTransport;
         this.#transportType = TransportType.MWP;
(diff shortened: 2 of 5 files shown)


packages/connect-multichain/src/multichain/index.ts
```diff
@@ -129,9 +129,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   }
 
   get transportType(): TransportType {
-    return this.#transport instanceof MWPTransport
-      ? TransportType.MWP
-      : TransportType.Browser;
+    return this.#transportType ?? TransportType.Browser;
```
jiexi: could we just invert the condition to `this.#transport instanceof DefaultTransport` instead? I'm not too keen on having manage the `this.#transportType` value if we don't have to

packages/connect-multichain/src/multichain/transports/mwp/KeyManager.ts
```diff
@@ -4,38 +4,49 @@ import type {
...
+ * Creates an {@link IKeyManager} backed by the `eciesjs` library.
+ *
+ * The factory dynamically imports `eciesjs` so the heavy crypto dependency is
+ * only loaded when MWP transport is actually used. The returned object closes
+ * over the imported symbols, allowing synchronous methods like
+ * `generateKeyPair` and `validatePeerKey` to work without a second await.
+ *
+ * @returns A ready-to-use key manager instance.
+ */
+export async function createKeyManager(): Promise<IKeyManager> {
```
jiexi, line 18: i like how you were able to encapsulate the dynamic importing complexity entirely into this file

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -360,32 +361,52 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   }
 
   async #createDappClient(): Promise<DappClient> {
+    const [mwpCore, { DappClient: DappClientClass }, { createKeyManager }] =
+      await Promise.all([
+        import('@metamask/mobile-wallet-protocol-core'),
+        import('@metamask/mobile-wallet-protocol-dapp-client'),
+        import('./transports/mwp/KeyManager'),
```
jiexi, line 374: does KeyManager need to be lazy loaded?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -584,7 +609,10 @@ export class MetaMaskConnectMultichain extends MultichainCore {
           resolve();
         })
         .catch(async (error) => {
-          if (error instanceof ProtocolError) {
+          if (
+            this.#mwpModules &&
+            error instanceof this.#mwpModules.ProtocolError
+          ) {
```
jiexi: thoughts on just lazy import mwp-core again in here to get access to the error classes rather than setting the error classes on the #mwpModules property on the MultichainClient instance?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -360,32 +361,52 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
     }
-    // Only setup MWP if it is not already mwp
     const { adapter: kvstore } = this.options.storage;
     const dappClient = await this.#createDappClient();
     this.#dappClient = dappClient;
-    const apiTransport = new MWPTransport(dappClient, kvstore);
+    const { MWPTransport: MWPTransportClass } = await import(
+      './transports/mwp'
+    );
+    const apiTransport = new MWPTransportClass(dappClient, kvstore);
```
jiexi:
```suggestion
    const { MWPTransport } = await import(
      './transports/mwp'
    );
    const apiTransport = new MWPTransport(dappClient, kvstore);
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -602,7 +630,10 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   async #setupDefaultTransport(
     options: { persist?: boolean } = { persist: true },
   ): Promise<DefaultTransport> {
-    if (this.#transport instanceof DefaultTransport) {
+    if (
+      this.#transportType === TransportType.Browser &&
+      this.#transport instanceof DefaultTransport
```
jiexi: why the double check here?

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/244/commits/50655140aefc51946648b0207d908cd91e4e370d

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: makes sense to me
https://github.com/MetaMask/connect-monorepo/pull/244/commits/50655140aefc51946648b0207d908cd91e4e370d

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: according to ticket spec, yes, it's intentional

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/244/commits/b6453226d8bf51c25a13cdf01ec7c6fb11e26998

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: The thing here is `#transportType` is used in too many places where `instanceof` won't work

Examples of this would be
- Reading the stored transport type before any instance exists, the public `transportType` getter that feeds into `RequestRouter/analytics`, 
- the `#setupMWP` early return (which would force an eager import and break lazy-loading
- the disconnect logic where the transport may already be partially torn down

This isn't just tracking what the transport is, it's tracking what it should be in states where the instance doesn't exist yet. The one spot where it is redundant is the `#setupDefaultTransport` guard, but we removed this in https://github.com/MetaMask/connect-monorepo/pull/244/commits/b6453226d8bf51c25a13cdf01ec7c6fb11e26998

packages/connect-multichain/CHANGELOG.md
```diff
@@ -11,18 +11,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Fix platform detection for Hermes-based React Native apps: `isReactNative()` now checks `global.navigator.product` before `window.navigator.product`, so modern RN environments where `window` is undefined correctly resolve to `PlatformType.ReactNative` instead of `PlatformType.NonBrowser`. This restores analytics for all React Native consumers — analytics were silently disabled because `#setupAnalytics()` only calls `analytics.enable()` for browser and `ReactNative` platforms ([#238](https://github.com/MetaMask/connect-monorepo/pull/238))
 
+## [0.11.1]
+
+### Changed
+
+- chore: align sub-package licenses with root ConsenSys 2022 license ([#241](https://github.com/MetaMask/connect-monorepo/pull/241))
+- chore: turborepo ([#239](https://github.com/MetaMask/connect-monorepo/pull/239))
```
adonesky1, line 32: these `### Fixed` entries got pulled up into `[Unreleased]` when the `## [0.11.1]` heading moved down — but #243 and #238 already shipped in 0.11.1, so they'd get re-announced in the next release notes. suggestion below keeps the new `### Changed` bullet under `[Unreleased]` and puts the two Fixed items back under `0.11.1`.

```suggestion
## [0.11.1]

### Changed

- chore: align sub-package licenses with root ConsenSys 2022 license ([#241](https://github.com/MetaMask/connect-monorepo/pull/241))
- chore: turborepo ([#239](https://github.com/MetaMask/connect-monorepo/pull/239))

### Fixed

- `MWPTransport.connect()` now accepts and forwards a `forceRequest` option. When `true`, `onResumeSuccess` skips the `isSameScopesAndAccounts` check and unconditionally sends `wallet_createSession`, allowing consumers to re-prompt account selection on an existing MWP session. Previously `forceRequest` was silently ignored, causing `wallet_requestPermissions` to no-op on mobile. ([#243](https://github.com/MetaMask/connect-monorepo/pull/243))
- Fix platform detection for Hermes-based React Native apps: `isReactNative()` now checks `global.navigator.product` before `window.navigator.product`, so modern RN environments where `window` is undefined correctly resolve to `PlatformType.ReactNative` instead of `PlatformType.NonBrowser`. This restores analytics for all React Native consumers — analytics were silently disabled because `#setupAnalytics()` only calls `analytics.enable()` for browser and `ReactNative` platforms ([#238](https://github.com/MetaMask/connect-monorepo/pull/238))
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -131,9 +125,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   }
 
   get transportType(): TransportType {
-    return this.#transport instanceof MWPTransport
-      ? TransportType.MWP
-      : TransportType.Browser;
+    return this.#transportType ?? TransportType.Browser;
```
adonesky1: the shadow field can drift from `#transport` — any future site that sets `#transport` but forgets `#transportType` makes this getter report `Browser`, which lands in the `transport_type` analytics field for what's actually an MWP session. consider deriving it instead (keeps one source of truth and doesn't force an eager MWP import):

```suggestion
  get transportType(): TransportType {
    if (this.#transport instanceof DefaultTransport) return TransportType.Browser;
    if (this.#transport) return TransportType.MWP;
    return TransportType.UNKNOWN;
  }
```

if you want to keep the shadow field, at minimum the fallback should probably be `UNKNOWN` rather than `Browser` so uninitialized state doesn't masquerade as a browser session in analytics.

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -616,6 +627,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       this.#onTransportNotification.bind(this),
     );
     this.#transport = transport;
+    this.#transportType = TransportType.Browser;
```
adonesky1, line 630: nit/consistency: the guard at line 618 in this same method still uses `this.#transport instanceof DefaultTransport`. everywhere else in this file was converted to `this.#transportType === TransportType.Browser` — worth flipping this one too so the identity check is uniform across the class.

adonesky1 (commented):
a few things:

1. changelog: the two `### Fixed` bullets (#243, #238) got pulled into `[Unreleased]` when `## [0.11.1]` moved — they already shipped. suggestion inline puts them back.
2. `transportType` getter now reads from a shadow field with a `Browser` fallback; easy to drift from `#transport` and silently mislabel analytics.
3. `#setupDefaultTransport` still does `instanceof DefaultTransport` while everywhere else switched to `#transportType === ...` — flip for consistency.

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/244/commits/3530e15632419f760a0c31b97d014c33cc618739

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/244/commits/3530e15632419f760a0c31b97d014c33cc618739

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -131,9 +125,13 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
-    return this.#transport instanceof MWPTransport
-      ? TransportType.MWP
-      : TransportType.Browser;
+    if (this.#transport instanceof DefaultTransport) {
+      return TransportType.Browser;
+    }
+    if (this.#transport) {
+      return TransportType.MWP;
+    }
+    return TransportType.UNKNOWN;
```
jiexi: since we have to manually maintain #transportType now. is this getter useful?

jiexi:
Validated that lazy loading is working in browser playground

packages/connect-multichain/src/multichain/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/244/commits/035a9a5aa95c63db1ed4d02bd91aadb7968f4c9f

adonesky1 (approved):

PR #245, opened 2026-03-31
fix: preserve selected chain on wallet_requestPermissions re-prompt

adonesky1:
## Summary

- **Preserve the user's current chain on `wallet_requestPermissions`** — `connect()` treats `chainIds[0]` as `#pendingPreferredChainId`, which wins over the cached chain when it's still permitted. Passing `[DEFAULT_CHAIN_ID]` (current behavior after revert) snaps back to mainnet on every re-prompt even when the user switched to e.g. Polygon. This puts `this.#provider.selectedChainId` first when it's still in the session's permitted set, so account-only re-prompts don't cause an unexpected `chainChanged`.
- **Fix analytics `scope` format** — uses `hexToNumber(preferred)` for correct decimal CAIP-2 format (`eip155:137`) instead of hex (`eip155:0x89`), matching `switchChain` and other handlers.
- **Add test** — verifies `wallet_requestPermissions` keeps the cached chain (`0x89` / Polygon) first in the scopes passed to `core.connect` when it's still in the multi-chain session.

## Test plan

- [ ] Run `yarn test` in `packages/connect-evm`
- [ ] Manual: connect on Polygon, call `wallet_requestPermissions`, verify chain stays on Polygon after account selection

### DIFF
--- a/packages/connect-evm/src/connect.test.ts
+++ b/packages/connect-evm/src/connect.test.ts
@@ -512,6 +512,46 @@ describe('MetamaskConnectEVM', () => {
       const forceRequestArg = secondConnectCall[3];
       expect(forceRequestArg).toBe(true);
     });
+
+    it('wallet_requestPermissions keeps cached chain first in chainIds when still permitted', async () => {
+      const mockCore = createMockCore();
+      mockCore.storage.adapter.get.mockResolvedValue(JSON.stringify('0x89'));
+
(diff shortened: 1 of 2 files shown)


wenfix:
added `no changelog` to this one: let's set the changelog in the parent PR.

wenfix (approved):
Confirmed it works! LGTM 🔥

PR #253, opened 2026-04-02
feat: Version enforcement using build and runtime checks

jiexi:
## Explanation

Adds runtime checks via:
1. Exposes the multichain-client package's version via MultichainClient.version
2. Adds a check to the MultichainClient.create() function that checks the singleton's version against the MultichainClient's version. If they don't match, there is a console warning, but we attempt to continue anyway
3. Adds a check to the ecosystem clients to check if the returned MultichainClient instance's version satisfies the expected peerDep version. It is not satisfied, there is a console warning, but we attempt to continue anyway

Adds build time checks via:
1. PeerDep on multichain-client from the ecosytem clients

## References

Fixes: https://consensyssoftware.atlassian.net/browse/WAPI-1262
See: https://consensys.slack.com/archives/C08SWDEN2GG/p1774566645252829

## Checklist


packages/connect-evm/src/connect.ts
```diff
@@ -1045,8 +1047,7 @@ export async function createEVMClient(
         supportedNetworks: supportedNetworksCaipChainId,
       },
       analytics: {
-        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-        integrationType: options.analytics?.integrationType || 'direct',
+        integrationType: options.analytics?.integrationType ?? 'direct',
```
jiexi: undo this

jiexi:
@SocketSecurity ignore npm/@metamask/connect-multichain@0.15.0

this is us. `fetch` makes sense

packages/connect-evm/package.json
```diff
@@ -70,6 +72,9 @@
     "typescript": "^5.9.3",
     "vitest": "^3.1.2"
   },
+  "peerDependencies": {
+    "@metamask/connect-multichain": "^0.0.0"
```
adonesky1: Is this intentional?

`^0.0.0` actually means `<0.0.1`, so `0.15.0` (current actual version) never matches it — everyone's going to get an unmet-peer warning on install. switch this to `workspace:^` and Yarn will publish `^0.15.0`?

packages/connect-solana/package.json
```diff
@@ -75,6 +77,9 @@
     "typescript": "^5.9.3",
     "vitest": "^3.1.2"
   },
+  "peerDependencies": {
+    "@metamask/connect-multichain": "^0.0.0"
```
adonesky1: same thing [as above](https://github.com/MetaMask/connect-monorepo/pull/253/changes#r3337288395):
> Is this intentional?

> ^0.0.0 actually means <0.0.1, so 0.15.0 (current actual version) never matches it — everyone's going to get an unmet-peer warning on install. switch this to workspace:^ and Yarn will publish ^0.15.0?

packages/connect-evm/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Validate `@metamask/connect-multichain` peer version at runtime and warn on mismatch ([#253](https://github.com/MetaMask/connect-monorepo/pull/253))
+
+### Changed
+
+- Move `@metamask/connect-multichain` from `dependencies` to `peerDependencies` ([#253](https://github.com/MetaMask/connect-monorepo/pull/253))
```
adonesky1:
```suggestion
- **BREAKING:** `@metamask/connect-multichain` is now a peer dependency.
  Add it to your own `dependencies` (e.g. `npm install @metamask/connect-multichain`)
  — it is no longer installed transitively.
```

packages/connect-solana/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Validate `@metamask/connect-multichain` peer version at runtime and warn on mismatch ([#253](https://github.com/MetaMask/connect-monorepo/pull/253))
+
+### Changed
+
+- Move `@metamask/connect-multichain` from `dependencies` to `peerDependencies` ([#253](https://github.com/MetaMask/connect-monorepo/pull/253))
```
adonesky1:
```suggestion
- **BREAKING:** `@metamask/connect-multichain` is now a peer dependency.
  Add it to your own `dependencies` (e.g. `npm install @metamask/connect-multichain`)
  — it is no longer installed transitively.
```

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -260,6 +257,16 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       | undefined;
     if (existing) {
       const instance = await existing;
+
+      if (instance.version !== getVersion()) {
```
adonesky1, line 262: the strict `!==` fires for any version diff, including compatible patches (0.15.0 vs 0.15.1) that deduped totally fine. the ecosystem checks use `satisfies()` against a range — could we make this consistent, e.g. only warn when major/minor differ, so a harmless patch bump doesn't tell people their setup is "NOT supported"?

packages/connect-evm/package.json
jiexi: yes. This value gets updated before being published via each package's tsup config.

The value here ideally would be `^workspace:`, but that isn't supported.

packages/connect-multichain/src/multichain/index.ts
jiexi: a patch bump may include an important fix though. Being totally silent with a version mismatch feels like the wrong approach IMO

packages/connect-evm/package.json
adonesky1: digging in more **with machine god** — this is worse than just the install warning. `^0.0.0` actually breaks the runtime check this PR adds.

`tsup.config.ts` computes the build define from the *source* manifest value:

```ts
resolveWorkspaceRange(pkg.peerDependencies['@metamask/connect-multichain'], multichainPkg.version)
```

`resolveWorkspaceRange` only rewrites `workspace:`-prefixed ranges, so `^0.0.0` passes straight through → `__CONNECT_MULTICHAIN_PEER_VERSION_RANGE__ = "^0.0.0"`. then at runtime:

```ts
satisfies(core.version, "^0.0.0") // satisfies("0.15.0", "^0.0.0") -> false
```

`^0.0.0` means `>=0.0.0 <0.0.1`, so a real version never satisfies it and the `console.warn` fires for *everyone*, even a perfectly deduped install. the check is inverted.

on "gets updated before publish" — that doesn't cover the build define. it's baked from the source `^0.0.0` at build time, and Yarn doesn't rewrite a literal `^0.0.0` at pack time (only `workspace:` ranges). so preview builds and local builds warn unconditionally.

`workspace:^` fixes all of it: Yarn publishes `^0.15.0` (no install warning), `resolveWorkspaceRange` resolves it to `^0.15.0`, and `satisfies("0.15.0", "^0.15.0")` → true. the helper was clearly written for `workspace:^` anyway.

re "`^workspace:` isn't supported" — that ordering isn't, but `workspace:^` is the correct form and *is* supported in `peerDependencies` (Yarn replaces it with `^<version>` at publish time: [docs](https://yarnpkg.com/features/workspaces), [berry#3386](https://github.com/yarnpkg/berry/pull/3386)). you're already using it in `devDependencies` here, and `resolveWorkspaceRange` has an explicit branch for the `^` case.

packages/connect-multichain/src/multichain/index.ts
adonesky1: Yeah you're right! NVM ✅

scripts/resolve-workspace-range.test.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable n/no-extraneous-import -- vitest is a root devDependency */
```
adonesky1: So this test doesn't actually run anywhere right now. CI runs `yarn turbo run test`, which only covers the workspaces (`packages/*`, etc), and `scripts/` isn't one. the root `test` script does call `yarn test:scripts`, but `test:scripts` isn't defined anywhere, so that path's dead too.

so `resolveWorkspaceRange` has no coverage in CI. easiest fix is probably a `test:scripts` script (`vitest run scripts`) wired into the test job, or move this into a workspace turbo already picks up.

packages/connect-evm/src/connect.ts
```diff
@@ -1203,6 +1205,21 @@ export async function createEVMClient(
...
 
+    const multichainClientPeerRange =
+      typeof __CONNECT_MULTICHAIN_PEER_VERSION_RANGE__ === 'undefined'
+        ? 'unknown'
+        : __CONNECT_MULTICHAIN_PEER_VERSION_RANGE__;
+
+    if (
+      multichainClientPeerRange !== 'unknown' &&
+      multichainClientPeerRange !== '' &&
+      !satisfies(core.version, multichainClientPeerRange)
```
adonesky1, line 1216: can we get a test on this branch? the singleton mismatch path is covered in `init.test.ts`, but this `satisfies(core.version, …)` warn has none — and it's the exact path that was inverted by the `^0.0.0` issue. a matching-version → no-warn / mismatch → warn pair here (plus the same in connect-solana) would've caught that directly and guards against it regressing.

jiexi: https://github.com/MetaMask/connect-monorepo/pull/253/commits/446b0b021f49de18318e36cf0e832d4b28865215

scripts/resolve-workspace-range.test.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/253/commits/0948b14f672735d48df2ae4be962d557648396c5

adonesky1: the `test:scripts` script needed a bit more wiring to actually run in CI, so I pushed c55b53d:

- added the `scripts/vitest.config.ts` it points at — that file wasn't in the branch, so `test:scripts` errored at startup. scoped it to `scripts/*.test.ts` so it doesn't pick up the jest-based `create-package` suite.
- added `yarn test:scripts` to the CI test job, since CI runs `yarn turbo run test` (workspaces only) and wasn't invoking it.

adonesky1 (approved):

PR #256, opened 2026-04-07
feat: add anon_id from dapp to connection metadata for cross-side analytics correlation

ffmcgee725:
## Explanation

Adds `anon_id` to the V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events — matching the cross-side correlation that SDK v1 already provides via `OriginatorInfo.anonId`.

- Added optional `analytics.anon_id` field to `ConnectionRequest.metadata` type
- Cached the `anonId` (from `storage.getAnonId()`) on the SDK singleton during init
- Introduced `#buildConnectionMetadata()` helper to DRY up the 4 metadata construction sites (`#renderInstallModalAsync`, `#headlessConnect`, `#deeplinkConnect`, `#openConnectDeeplinkIfNeeded`)
- The `analytics` block is only included when `anonId` is available, keeping the payload backward-compatible

## Related

- Wallet-side counterpart: [MetaMask/metamask-mobile#28470](https://github.com/MetaMask/metamask-mobile/pull/28470)
- [WAPI-1380](https://consensyssoftware.atlassian.net/browse/WAPI-1380)

## References

* Fixes https://consensyssoftware.atlassian.net/browse/WAPI-1382

## Checklist

### DIFF
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Include `analytics.anon_id` in V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events ([#256](https://github.com/MetaMask/connect-monorepo/pull/256))
+- Include `analytics.remote_session_id` in V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events ([#256](https://github.com/MetaMask/connect-monorepo/pull/256))
 
 ## [0.11.1]
 
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Include `analytics.anon_id` in V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events ([#256](https://github.com/MetaMask/connect-monorepo/pull/256))
+
 ## [0.11.1]
 
 ### Changed
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -357,7 +357,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
       sdk: { version: getVersion(), platform: getPlatformType() },
     };
     if (this.#anonId) {
-      metadata.analytics = { anon_id: this.#anonId };
+      metadata.analytics = { remote_session_id: this.#anonId };
     }
     return metadata;
   }
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -365,7 +365,6 @@ export class MetaMaskConnectMultichain extends MultichainCore {
   async #init(): Promise<void> {
     try {
       await this.#setupAnalytics();
-      this.#anonId ??= await this.storage.getAnonId();
       await this.#setupTransport();
     } catch (error) {
       await this.storage.removeTransport();
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -94,6 +94,8 @@ export class MetaMaskConnectMultichain extends MultichainCore {
 
   #listener: (() => void | Promise<void>) | undefined;
 
+  #anonId: string | undefined;
+
   get status(): ConnectionStatus {
     return this._status;
   }
@@ -242,6 +244,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
 
     const dappId = getDappId(this.options.dapp);
     const anonId = await this.storage.getAnonId();
+    this.#anonId = anonId;
 
     const { integrationType } = this.options.analytics ?? {
       integrationType: '',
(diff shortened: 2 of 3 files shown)


packages/connect-multichain/src/multichain/index.ts
```diff
@@ -348,9 +351,21 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     }
   }
 
+  #buildConnectionMetadata(): ConnectionRequest['metadata'] {
+    const metadata: ConnectionRequest['metadata'] = {
+      dapp: this.options.dapp,
+      sdk: { version: getVersion(), platform: getPlatformType() },
+    };
+    if (this.#anonId) {
```
adonesky1, line 359: under what conditions would we expect this to be falsy?

ffmcgee725: The guard is purely defensive: avoid sending `analytics: { anon_id: "" }` if for some reason `getAnonId()` throws.

If you're confident `getAnonId()` never returns empty, the guard could be dropped — but it's cheap insurance and keeps the payload clean.

adonesky1 (approved):
LGTM

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -348,9 +351,21 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
+      metadata.analytics = { anon_id: this.#anonId };
+    }
+    return metadata;
+  }
+
   async #init(): Promise<void> {
     try {
       await this.#setupAnalytics();
+      this.#anonId ??= await this.storage.getAnonId();
```
jiexi: shouldn't this be have been set in `this.#setupAnalytics()` above already if it needed to be set?

packages/connect-multichain/src/multichain/index.ts
jiexi: if there is no anonId, i think that implies that analytics was not setup. Analytics isn't setup because we don't want to track events for that environment. So I think it's probably fine to remove this guard since it should only ever be falsy when we aren't going to track the event anyway?

adonesky1: at least currently I don't believe there are any contexts/environments in which we're not emitting tracking events for dapp side code?

packages/connect-multichain/src/multichain/index.ts
adonesky1: Yep that looks like the case to me as well 🤔

adonesky1 (dismissed):
lets address Jiexi's [comment before merging](https://github.com/MetaMask/connect-monorepo/pull/256/changes#r3053672006)

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -94,6 +94,8 @@ export class MetaMaskConnectMultichain extends MultichainCore {
 
   #listener: (() => void | Promise<void>) | undefined;
 
+  #anonId: string | undefined;
```
jiexi, line 97: i understand that we're adding this property to avoid an async call to `this.storage.getAnonId()` in subsequent usage, but i'm not sure it's worth having to add a property to this class that is otherwise unrelated to anything else going on in the class. this is a nit though

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: `#setupAnalytics` has an early return if platform is neither browser nor React Native

```ts
if (!isBrowser && !isReactNative)
```

so this fallback exists to cover to make sure `#anonId` is still populated for other environments (e.g., for Node.js/server environments) so the metadata at lines 359-360 will have an `anon_id` on subsequent calls.

So the line isn't truly redundant — it's a fallback for non-browser, non-React-Native platforms.

The question is, do we want to keep track of `anon_id` in these environments ? I don't remember if analytics can be toggled on for these or not.

packages/connect-multichain/src/multichain/index.ts
ffmcgee725: I would prefer keeping it as such, since removing it would make `#buildConnectionMetadata()` method call need to be asynchronous, which is currently called in 4 places across this class.

packages/connect-multichain/src/multichain/index.ts
jiexi: i don't believe we do want to track them outside of browser and react-native environments based on the existing logic in `this.#setupAnalytics()`. So I still believe that this line isn't needed

packages/connect-multichain/src/multichain/index.ts
jiexi: Fair. always a balance

packages/connect-multichain/src/multichain/index.ts
jiexi: we can just leave this

packages/connect-multichain/src/multichain/index.ts
adonesky1: The other environment being node?

adonesky1: Yeah so I think @jiexi is right, if setupAnalytics is not being called it means we aren't going to be doing any analytics in that environment. I guess that case is just for node right now. In any case it doesn't make sense to store a class variable that is only related to analytics in contexts where we aren't actually going to be doing analytics.

ffmcgee725: https://github.com/MetaMask/connect-monorepo/pull/256/commits/05d6943447e55a8bcc7dc26f683040aff86670c2

ffmcgee725:
@adonesky1 Jiexi's comment was addressed https://github.com/MetaMask/connect-monorepo/pull/256/commits/05d6943447e55a8bcc7dc26f683040aff86670c2

packages/connect-multichain/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Include `analytics.anon_id` in V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events ([#256](https://github.com/MetaMask/connect-monorepo/pull/256))
```
jiexi:
```suggestion
- Include `analytics.remote_session_id` in V2 connection metadata so the wallet can correlate dapp-side and wallet-side analytics events ([#256](https://github.com/MetaMask/connect-monorepo/pull/256))
```

jiexi (approved):

PR #265, opened 2026-04-15
fix: ensure connect-evm client is fully initialized before returning client instance in createEVMClient

jiexi:
## Explanation

Fixes a bug where `eth_accounts` was resolving with `[]` immediately after a successful call to `MultichainClient.connect()`. This seems to be the result of a race condition where the MultichainClient has not yet cached a value for `eth_accounts`, resulting in quick enough calls to `request()` resulting in the default value of `[]`

This PR fixes that by ensuring `createEVMClient()` doesn't return until the underlying instance is fully initialized.

## Checklist


adonesky1 (approved):
LGTM

PR #268, opened 2026-04-16
fix: prevent CSP connect-src violation from QR code image fetch

wenfix:
## Problem

When the install modal renders the QR code, `qr-code-styling` was triggering a `connect-src` CSP violation on host pages. The root cause: Stencil inlines `fox.svg` as a `data:image/svg+xml;base64,...` URI, which gets passed as the `image` option to `qr-code-styling`. That library defaults `imageOptions.saveAsBlob` to `true`, which causes it to do:

```js
xhr.open("GET", "data:image/svg+xml;base64,...");
xhr.responseType = "blob";
xhr.send();
```

Browsers treat XHR on `data:` URIs as network requests and check them against `connect-src`. Since host pages have no reason to include `data:` in their `connect-src`, this violated their CSP silently in production.

## Fix

Set `saveAsBlob: false` in the QR code `imageOptions`. Since we only ever append the QR code to the DOM (never export it), the blob conversion is unnecessary work. Also removed the `crossOrigin: 'anonymous'` setting, which is meaningless for a data URI.

```diff
 imageOptions: {
   hideBackgroundDots: true,
-  crossOrigin: 'anonymous',
+  crossOrigin: undefined,
   imageSize: 0.3,
+  saveAsBlob: false,
 },
```

## Browser Playground

Added a restrictive `Content-Security-Policy` meta tag to the browser playground so this class of violation is caught locally during manual testing, rather than discovered in production. The policy mirrors what a typical host page would enforce — notably `connect-src` does **not** include `data:`.

## Notes for consumers

Host pages integrating this library still need the following in their CSP:

```
connect-src wss://mm-sdk-relay.api.cx.metamask.io;
```

This is unavoidable — it's the WebSocket relay and cannot be proxied or deferred from within the library.

## Test plan

- [ ] Run `yarn start` in `browser-playground`, open DevTools console, click Connect — no CSP violations should appear when the QR modal renders
- [ ] To verify the fix is necessary: temporarily revert `saveAsBlob: false`, repeat — a `connect-src` violation for the `data:image/svg+xml` URI should appear in the console

### DIFF
--- a/playground/browser-playground/public/index.html
+++ b/playground/browser-playground/public/index.html
@@ -13,8 +13,9 @@
       content="default-src 'self';
                script-src 'self' 'unsafe-inline' 'unsafe-eval';
                style-src 'self' 'unsafe-inline';
-               img-src 'self' data: blob:;
-               connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://*.infura.io;
+               img-src 'self' data:;
+               connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://mm-sdk-analytics.api.cx.metamask.io https://*.infura.io;
+               frame-src https://fwd.metamask.io;
                font-src 'self';"
     />
   </head>
--- a/playground/browser-playground/public/index.html
+++ b/playground/browser-playground/public/index.html
@@ -8,6 +8,15 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <title>Multichain Test Dapp</title>
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self';
+               script-src 'self' 'unsafe-inline' 'unsafe-eval';
+               style-src 'self' 'unsafe-inline';
+               img-src 'self' data: blob:;
+               connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://*.infura.io;
+               font-src 'self';"
+    />
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ img-src data:;
 
 - `wss://mm-sdk-relay.api.cx.metamask.io` — the WebSocket URL of the MetaMask relay used for remote connections (mobile, no-extension, etc.). Unavoidable — the relay cannot be proxied or deferred from within the library, and remote connections will fail without it.
 - `https://mm-sdk-analytics.api.cx.metamask.io` — the telemetry endpoint used by `@metamask/analytics`. Required for the connection lifecycle events the SDK emits by default. You can override the host via the `METAMASK_ANALYTICS_ENDPOINT` env var at build time, but some analytics endpoint must be reachable.
-- `img-src data:` — the install/QR-code modal in `@metamask/multichain-ui` embeds the MetaMask fox SVG as a `data:` URI inside the generated QR code. Without this, the QR code will fail to render the center logo.
+- `img-src data:` — the install/QR-code modal in `@metamask/multichain-ui` embeds the MetaMask fox SVG as a `data:` URI inside the generated QR code. Without this, the QR code will fail to render entirely.
 
 ### Also consider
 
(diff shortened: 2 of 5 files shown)


playground/browser-playground/public/index.html
```diff
@@ -8,6 +8,15 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <title>Multichain Test Dapp</title>
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self';
+               script-src 'self' 'unsafe-inline' 'unsafe-eval';
+               style-src 'self' 'unsafe-inline';
+               img-src 'self' data: blob:;
+               connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://*.infura.io;
```
jiexi: why is this infura.io and not metamask.io?

adonesky1:
```suggestion
connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://mm-sdk-analytics.api.cx.metamask.io https://*.infura.io;
frame-src https://fwd.metamask.io;
```

Claude believes this CSP will break two SDK flows as soon as you actually use the playground:

1. analytics (`@metamask/analytics`) POSTs to https://mm-sdk-analytics.api.cx.metamask.io/ — auto-enabled in browser via setupAnalytics in connect-multichain/src/multichain/index.ts:262, fires on every track() in connect-evm.

2. MetaMaskOnboarding.startOnboarding() (ModalFactory.ts:147) injects an `<iframe src="https://fwd.metamask.io">`, which falls back to default-src 'self' and gets blocked — the "install extension" flow silently breaks.

adonesky1: https://github.com/user-attachments/assets/25d981b2-e0e3-4919-8c4e-548d54962b74

adonesky1: I'm also not seeing the QR code show on this branch 

Image

playground/browser-playground/public/index.html
```diff
@@ -8,6 +8,15 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <title>Multichain Test Dapp</title>
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self';
+               script-src 'self' 'unsafe-inline' 'unsafe-eval';
+               style-src 'self' 'unsafe-inline';
+               img-src 'self' data: blob:;
```
jiexi: is `blob` needed here? I would imagine not since this PR configures Stencil not to create blobs and instead just use the inlined `data:` uri

jiexi: just checked. seems like this is the case

playground/browser-playground/public/index.html
wenfix: Both are needed for different purposes: `wss://mm-sdk-relay.api.cx.metamask.io` is the WebSocket relay and `https://*.infura.io` is for RPC calls

wenfix: Done — added both `mm-sdk-analytics` and `frame-src` in 1c97bbe7.

playground/browser-playground/public/index.html
wenfix: Correct, removed in 1c97bbe7.

playground/browser-playground/public/index.html
wenfix: I'm seeing the QR code fix work. Did you re-build the packages? What browser are you using?

ffmcgee725: After checking out this branch, made sure to 

```bash
yarn && yarn build
yarn workspace @metamask/browser-playground start
```

And opening up playground + calling connect on a browser with no extension installed, I get the QR code to pop

Screenshot 2026-04-17 at 12 51 14

jiexi: righhhhht. Thank you

adonesky1: > I'm seeing the QR code fix work. Did you re-build the packages? What browser are you using?

I'm confused... were you seeing it work before the changes I suggested or after?

adonesky1 (dismissed):
not sure about why all the lint changes on `mm-install-modal.tsx`

but otherwise LGTM

jiexi (approved):

PR #269, opened 2026-04-16
fix: ensure EIP-1193 provider properties are updated before emitting `connect` event

jiexi:
## Explanation

Previously, the EvmClient and EIP-1193 Provider would emit `connect` events before the EIP-1193 Provider properties were updated. These values were also used as the cache for `eth_chainId` and `eth_accounts` requests. This resulted in empty values being returned to the caller when making requests to either method inside the event handler for a `connect` event.

This PR fixes that by ensuring the EIP-1193 provider's property values are updated before emitting `connect`

## Checklist

### DIFF
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -802,8 +802,8 @@ export class MetamaskConnectEVM {
     }
     logger('handler: chainChanged', { chainId });
     this.#provider.selectedChainId = chainId;
-    this.#eventHandlers?.chainChanged?.(chainId);
     this.#provider.emit('chainChanged', chainId);
+    this.#eventHandlers?.chainChanged?.(chainId);
   }
 
   /**
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -3,6 +3,7 @@
 import { analytics } from '@metamask/analytics';
 import { parseScopeString } from '@metamask/chain-agnostic-permission';
 import type {
+  ConnectionStatus,
   MultichainCore,
   MultichainOptions,
   Scope,
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -3,7 +3,6 @@
 import { analytics } from '@metamask/analytics';
 import { parseScopeString } from '@metamask/chain-agnostic-permission';
 import type {
-  ConnectionStatus,
   MultichainCore,
   MultichainOptions,
   Scope,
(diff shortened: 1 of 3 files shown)


packages/connect-evm/src/connect.ts
```diff
@@ -844,37 +844,47 @@ export class MetamaskConnectEVM {
...
-    }
+    this.#status = 'connected';
+    this.#provider.emit('connect', data);
+    this.#eventHandlers?.connect?.(data);
+
+    this.#provider.emit('chainChanged', chainId);
+    this.#eventHandlers?.chainChanged?.(chainId);
+
+    this.#provider.emit('accountsChanged', accounts);
+    this.#eventHandlers?.accountsChanged?.(accounts);
```
adonesky1, line 864: right?
```suggestion
if (chainId !== prevChainId) {
  this.#provider.emit('chainChanged', chainId);
  this.#eventHandlers?.chainChanged?.(chainId);
}
if (!accountsEqual(accounts, prevAccounts)) {
  this.#provider.emit('accountsChanged', accounts);
  this.#eventHandlers?.accountsChanged?.(accounts);
}
```

packages/connect-evm/src/connect.ts
```diff
@@ -844,37 +844,47 @@ export class MetamaskConnectEVM {
...
-        this.#core.off('metamask_accountsChanged', onAccountsChanged);
-        this.#core.off('metamask_chainChanged', onChainChanged);
-      };
-    }
+    this.#status = 'connected';
+    this.#provider.emit('connect', data);
+    this.#eventHandlers?.connect?.(data);
+
+    this.#provider.emit('chainChanged', chainId);
+    this.#eventHandlers?.chainChanged?.(chainId);
```
adonesky1, line 861: total nit but is there some reason we emit before invoking the handler, and in `#onChainChanged` we do the opposite: https://github.com/MetaMask/connect-monorepo/blob/200cc1ae0c73c41498db0516519096d4b02fb458/packages/connect-evm/src/connect.ts#L805-L806

packages/connect-evm/src/connect.ts
jiexi: accounts will for sure not be equal already. chainId might, but i don't think it really matters if we emit chainChanged unnecessarily so I opted not to have a guard for either out of simplicity

packages/connect-evm/src/connect.ts
jiexi: no. no reason at all. I can make these consistent

jiexi: https://github.com/MetaMask/connect-monorepo/pull/269/commits/53a3c060c59fde7b0aa7ed295523fbaed224ca40

adonesky1 (approved):
LGTM

PR #270, opened 2026-04-16
fix: Update EvmClient.status to reflect actual instance status

jiexi:
## Explanation

Currently `EvmClient.status` returns the underlying `MultichainClient.status` value. This is problematic as the MultichainClient being connected does not necessarily mean that EVM accounts are permissioned. Additionally, the MultichainClient's connected status does not correctly indicated whether or not the EvmClient itself has internally reached a fully connected state.

This PR fixes that by making `EvmClient.status` reflect the actual internal status of the EvmClient instance. This is a breaking change to the API.

## Checklist

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING** `EvmClient.status` now reflects the actual internal status of the `EvmClient` instance instead of proxying the underlying `MultichainClient.status`. The return type changes from `ConnectionStatus` to `ConnectEvmStatus`. ([#270](https://github.com/MetaMask/connect-monorepo/pull/270))
+
 ## [0.10.0]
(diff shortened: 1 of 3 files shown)


adonesky1:
[nit] 

Two additional test cases we could consider adding now that `status` is part of the public contract:

1. **user disconnects from the wallet.** the current 'disconnected' test only covers user-initiated client.disconnect(). the wallet can also end the session (e.g. user revokes permissions in metamask), which hits #onSessionChanged → #onDisconnect at line 771. 

- **new test:** connect, emit wallet_sessionChanged with empty scopes, assert client.status === 'disconnected'.

2. **failed connect() resets status.** connect() sets #status = 'connecting' up front, and relies on the catch block at line 398 to reset it to 'disconnected' on error. that reset is easy to accidentally drop in a future refactor — and now that status is public, a bug there leaves consumers observing a stuck 'connecting' state with no recovery path.

- **new test:** make mockCore.connect reject, await the rejection from client.connect(...), assert client.status === 'disconnected'.

adonesky1 (dismissed):

adonesky1 (approved):
LGTM!

PR #275, opened 2026-04-21
feat: prefer injected solana provider

jiexi:
## Explanation

Prefer the injected solana provider now that we have passive listening for extension and IAB that ensures our other ecosystem clients will be notified when an injected solana provider initiates a wallet_createSession

This can be tested by:
1. Using browser playground with extension enabled
2. Connect to solana, be sure to include EVM networks
3. EVM and Solana cards should appear in the playground despite the Solana connection flow going through the injected Solana provider
4. ~~Check in the console.logs to ensure the injected provider was preferred~~ Add breakpoints to the `connect.ts` code to ensure the injected provider was announced first. The debug logger is broken currently, see [slack](https://consensys.slack.com/archives/C08SWDEN2GG/p1776802144246869). 

https://github.com/user-attachments/assets/e3da8893-32b3-4d62-8a27-0258481751da

## References

Fixes: https://consensyssoftware.atlassian.net/browse/WAPI-1083

## Checklist

### DIFF
--- a/packages/connect-solana/src/connect.ts
+++ b/packages/connect-solana/src/connect.ts
@@ -99,22 +99,36 @@ export async function createSolanaClient(
   const walletName = 'MetaMask';
 
   let hasRegisteredMmc = false;
-  let handleInitRegistration!: () => void;
+  let handledInitRegistration!: () => void;
   const initRegistrationHandledPromise = new Promise<void>((resolve) => {
-    handleInitRegistration = resolve;
+    handledInitRegistration = resolve;
   });
 
+  const registerWallet = async (): Promise<void> => {
+    if (hasRegisteredMmc) {
+      logger('MetaMask Connect is already registered. Skipping...');
+      return;
+    }
+
+    if (isMetamaskExtensionRegistered()) {
+      logger('MetaMask extension is already registered. Skipping...');
+      return;
+    }
+
+    hasRegisteredMmc = true;
+    await registerSolanaWalletStandard({ client, walletName });
+  };
+
   if (skipAutoRegister) {
-    handleInitRegistration();
+    handledInitRegistration();
   } else {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
     setTimeout(async () => {
-      if (isMetamaskExtensionRegistered()) {
-        logger('MetaMask extension is already registered. Skipping...');
-      } else {
-        await registerSolanaWalletStandard({ client, walletName });
-        hasRegisteredMmc = true;
+      try {
+        await registerWallet();
+      } finally {
+        handledInitRegistration();
       }
-      handleInitRegistration();
     }, 1000);
   }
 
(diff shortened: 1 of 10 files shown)


packages/connect-solana/src/connect.ts
```diff
@@ -91,20 +96,48 @@ export async function createSolanaClient(
...
+      return;
+    }
 
-  if (!skipAutoRegister) {
+    if (isMetamaskExtensionRegistered()) {
+      logger('MetaMask extension is already registered. Skipping...');
+      return;
+    }
+
+    hasRegisteredMmc = true;
```
adonesky1: should this come after `await registerSolanaWalletStandard({ client, walletName })` ?

packages/connect-solana/src/connect.ts
```diff
@@ -60,6 +61,10 @@ declare const __PACKAGE_VERSION__: string | undefined;
 export async function createSolanaClient(
   options: SolanaConnectOptions,
 ): Promise<SolanaClient> {
+  if (options.debug) {
+    enableDebug();
+  }
```
adonesky1: unrelated to core purpose of the PR yeah?

packages/connect-solana/src/connect.ts
```diff
@@ -91,20 +96,48 @@ export async function createSolanaClient(
 
   const client = core.provider;
 
-  const walletName = 'MetaMask Connect';
+  const walletName = 'MetaMask';
```
adonesky1, line 95: This is breaking no?

packages/connect-solana/src/connect.ts
```diff
@@ -91,20 +96,48 @@ export async function createSolanaClient(
...
+  } else {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    setTimeout(async () => {
+      try {
+        await registerWallet();
+      } finally {
+        handledInitRegistration();
+      }
+    }, 1000);
   }
```
adonesky1, line 130: Should we 
```suggestion
  // don't hand the client to the caller until the auto-register
  // decision has actually been made, so the wallet-standard registry is
  // consistent with the "auto-registered on creation" contract.
  await initRegistrationHandledPromise;
```

adonesky1: not sure if we definitely want to do this since it will mean the wallet doesn't show up as registered for 1 second

packages/connect-solana/src/connect.ts
jiexi: linter complains that this should be atomic, i.e. before giving up the execution loop. I agree that it should come after though. I'll change it and add a lint ignore

packages/connect-solana/src/connect.ts
jiexi: absolutely 100%. need changelog

packages/connect-solana/src/connect.ts
jiexi: yeah. good catch

jiexi: https://github.com/MetaMask/connect-monorepo/pull/275/commits/243a669d04c8e1c0d86fb55688a3c0a04d61ac9b

packages/connect-solana/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/275/commits/631e97779f5495bbfa101eda2255dce8e4774188

packages/connect-solana/src/connect.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/275/commits/515ded1a6772e07bcba56deca94143c154e7797a

adonesky1 (approved):
LGTM

PR #282, opened 2026-04-27
feat: ensure solana provider fully hydrated before resolve

jiexi:
## Explanation

In `connect-solana`, wait for the Solana wallet-standard provider to be fully hydrated before being registered via wallet standard events and before being returned in the result of `getWallet()`.

`getWallet()` also now returns a singleton Solana wallet-standard provider

## Checklist

### DIFF
--- a/packages/connect-solana/src/connect.ts
+++ b/packages/connect-solana/src/connect.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention -- __PACKAGE_VERSION__ is an esbuild define convention */
+import { parseScopeString } from '@metamask/chain-agnostic-permission';
 import {
   createMultichainClient,
   type Scope,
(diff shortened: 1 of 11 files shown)


ffmcgee725 (commented):
LGTM! But CI needs fixing ⚠️

packages/connect-solana/src/connect.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention -- __PACKAGE_VERSION__ is an esbuild define convention */
+import { parseScopeString } from '@metamask/chain-agnostic-permission';
```
adonesky1: [nit] is it worth adding this whole package for this?

adonesky1: I think most of the time this will get treeshaken by consumers but for those cases where its not I think this will pull in a ton of unnecessary transitive deps

adonesky1 (changes requested):
I think we should just recreate `parseScopeString` inline here to avoid the unnecessary imports

packages/connect-solana/src/connect.ts
jiexi: fair point. we should probably remove it from connect-evm as well then

adonesky1: yes... we should

adonesky1: will create a PR for that shortly here

adonesky1 (dismissed):

adonesky1 (approved):

PR #285, opened 2026-05-06
feat: send `sessionProperties: { eip1193-compatible': true }` on every `wallet_createSession` request issued by `connect-evm`

ffmcgee725:
## Summary

Sends `sessionProperties: { 'eip1193-compatible': true }` on every `wallet_createSession` request issued by `@metamask/connect-evm`. This lets the wallet distinguish EIP-1193-style connections established through `connect-evm` (which surfaces a familiar `window.ethereum`-like interface on top of the Multichain API) from pure Multichain API connections and other provider types like Solana Wallet Standard.

The wallet can then apply EVM-specific UX — most immediately, conditionally rendering the dapp connection bar's network picker only for connections that have this property — without inflicting that UX on non-EVM connections.

Jira: [WAPI-1482](https://consensyssoftware.atlassian.net/browse/WAPI-1482)

### `packages/connect-evm/src/constants.ts`

- Added `EIP1193_COMPATIBLE_SESSION_PROPERTY` constant (`'eip1193-compatible'`).
- Added `CONNECT_EVM_SESSION_PROPERTIES` constant (`{ 'eip1193-compatible': true }`) — the session properties object sent with every connection.

### `packages/connect-evm/src/connect.ts`

- In `MetamaskConnectEVM.connect()`, replaced the `undefined` third argument to `this.#core.connect()` with `CONNECT_EVM_SESSION_PROPERTIES`. The `connect-multichain` layer already supports `sessionProperties` end-to-end (`mergeRequestedSessionWithExisting()` and all transport `connect()` paths), so the property flows through to the `wallet_createSession` request payload without any changes to `connect-multichain`.

### `packages/connect-evm/src/connect.test.ts`

- Added a new `describe('sessionProperties', ...)` block with two tests:
  - `connect()` passes `{ 'eip1193-compatible': true }` to `core.connect`.
  - `wallet_requestPermissions` re-entry path also passes the same session properties.

### `packages/connect-evm/CHANGELOG.md`

- Added an `### Added` entry under `[Unreleased]` documenting the new session property and the rationale.

## Test plan

- [x] Unit tests pass (41/41) — `yarn workspace @metamask/connect-evm run test:unit`
- [x] Lint passes with no warnings — `npx eslint packages/connect-evm/src/**`
- [x] Build succeeds — `yarn workspace @metamask/connect-evm run build`
- [x] Changelog validates — `yarn workspace @metamask/connect-evm run changelog:validate`
- [ ] Manual smoke test: connect via `connect-evm` against an extension build that surfaces the new session property and verify it appears in the persisted CAIP-25 caveat (requires the downstream `chain-agnostic-permission` change [here](https://github.com/MetaMask/core/pull/8731))

### DIFF
--- a/packages/connect-evm/src/connect.test.ts
+++ b/packages/connect-evm/src/connect.test.ts
@@ -131,7 +131,7 @@ describe('MetamaskConnectEVM', () => {
       });
     });
 
-    it('passes { "eip1193-compatible": true } as sessionProperties on connect via wallet_requestPermissions', async () => {
+    it('passes { "eip1193-compatible": true } as sessionProperties when `wallet_requestPermissions` is called directly on provider', async () => {
       const mockCore = createMockCore();
       mockCore.storage.adapter.get.mockResolvedValue(JSON.stringify('0x1'));
       mockCore.connect.mockImplementation(async (): Promise<void> => {
--- a/packages/connect-evm/src/connect.test.ts
+++ b/packages/connect-evm/src/connect.test.ts
@@ -104,6 +104,71 @@ function createMockCore(): MockCore {
 }
 
 describe('MetamaskConnectEVM', () => {
+  describe('sessionProperties', () => {
+    it('passes { "eip1193-compatible": true } as sessionProperties on connect', async () => {
+      const mockCore = createMockCore();
+      mockCore.storage.adapter.get.mockResolvedValue(JSON.stringify('0x1'));
+      mockCore.connect.mockImplementation(async (): Promise<void> => {
+        const session: SessionData = {
+          sessionScopes: {
+            'eip155:1': {
+              methods: [],
+              notifications: [],
+              accounts: ['eip155:1:0x1234567890123456789012345678901234567890'],
+            },
+          },
+        };
+        mockCore.emit('wallet_sessionChanged', session);
+      });
+      const client = await MetamaskConnectEVM.create({ core: mockCore });
+
+      await client.connect({ chainIds: ['0x1'] });
+
+      expect(mockCore.connect).toHaveBeenCalledTimes(1);
+      const [, , sessionProperties] = mockCore.connect.mock.calls[0];
+      expect(sessionProperties).toStrictEqual({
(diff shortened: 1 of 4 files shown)


adonesky1:
From the final manual step:
> requires the downstream chain-agnostic-permission change above

I'm not seeing the PR/Change that this is referencing 🤔 I'm assuming we have to make a change there to make this an accepted session property IIRC

packages/connect-evm/src/connect.test.ts
```diff
@@ -104,6 +104,71 @@ function createMockCore(): MockCore {
...
+      await client.connect({ chainIds: ['0x1'] });
+
+      expect(mockCore.connect).toHaveBeenCalledTimes(1);
+      const [, , sessionProperties] = mockCore.connect.mock.calls[0];
+      expect(sessionProperties).toStrictEqual({
+        'eip1193-compatible': true,
+      });
+    });
+
+    it('passes { "eip1193-compatible": true } as sessionProperties on connect via wallet_requestPermissions', async () => {
```
adonesky1: [nit] this a bit confusing since its technically `wallet_createSession` where the session property is added
```suggestion
    it('passes { "eip1193-compatible": true } as sessionProperties when `wallet_requestPermissions` is called directly on provider', async () => {
```

adonesky1 (dismissed):
a few non blocking comments. Otherwise LGTM

ffmcgee725:
@adonesky1 I opened up that PR in core https://github.com/MetaMask/core/pull/8731

we need to insert this new session property [here](https://github.com/MetaMask/core/blob/ab58df0303b769becb1129959d5fb7ddcab01ada/packages/chain-agnostic-permission/src/scope/constants.ts#L100), otherwise calling `wallet_createSession` with this new session property will just throw [here](https://github.com/MetaMask/core/blob/ab58df0303b769becb1129959d5fb7ddcab01ada/packages/chain-agnostic-permission/src/caip25Permission.ts#L206)

wenfix (approved):

PR #287, opened 2026-05-07
fix: Fix `wallet_switchEthereumChain` to forward 'Unrecognized chain ID' error 

jiexi:
## Explanation

Previously, if the EIP-1193 provider was called with `wallet_switchEthereumChain` or the EvmClient's `switchChain()` was called without an add chain `chainConfiguration` fallback AND the chain did not exist on the wallet, the caller would get an error `"No chain configuration found."`. This was an incorrect and breaking change to the `Unrecognized chain ID` error expected by dapps for `wallet_switchEthereumChain` . 

This has been fixed by changing `switchChain()`/`wallet_switchEthereumChain` to correctly forward/re-throw a `Unrecognized chain ID` error if `chainConfiguration` is not provided.

## Checklist

### DIFF
--- a/packages/connect-evm/src/connect.test.ts
+++ b/packages/connect-evm/src/connect.test.ts
@@ -84,6 +84,7 @@ function createMockCore(): MockCore {
     disconnect: vi.fn().mockResolvedValue(undefined),
     connect: vi.fn().mockResolvedValue(undefined),
     invokeMethod: vi.fn().mockResolvedValue('0xsignature'),
+    openSimpleDeeplinkIfNeeded: vi.fn(),
     provider: {
       getSession: vi.fn().mockResolvedValue({ sessionScopes: {} }),
     },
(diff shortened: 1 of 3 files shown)


packages/connect-evm/src/connect.test.ts
```diff
@@ -84,6 +84,7 @@ function createMockCore(): MockCore {
     disconnect: vi.fn().mockResolvedValue(undefined),
     connect: vi.fn().mockResolvedValue(undefined),
     invokeMethod: vi.fn().mockResolvedValue('0xsignature'),
+    openSimpleDeeplinkIfNeeded: vi.fn(),
```
adonesky1, line 87: This is because `switchChain` calls this and we didn't previously need to mock it?

adonesky1 (approved):
Non-blocking question/comment!

packages/connect-evm/src/connect.test.ts
jiexi: tbh i didn't check

PR #290, opened 2026-05-11
feat(analytics): attach `failure_reason` to `_failed` events + fix `isRejectionError`

adonesky1:
Sibling PRs from the same WAPI-1253 spike (independent — can land in any order):

- #292 — fix `isRejectionError` (unwrap `RPCInvokeMethodErr`, stop treating `4100` as a user rejection, tighten the "user" message heuristic).
- #293 — adds `error_code` and `error_message_sample` companions to `failure_reason` so the `unknown` bucket stays triageable without per-code schema migrations. Stacked on this PR.

(The previously-paired #291 — _"dedupe EVM shim analytics tracking"_ — was closed: source review showed the EVM-intercepted-method paths never reach `RequestRouter`, so there was no duplication to remove. See the close comment on #291 for details.)

> **Note on the relationship with #292:** the new `wallet_unauthorized` classifier bucket in this PR only sees traffic once #292 lands and stops swallowing `4100` as `_rejected`. Until then it's forward-compatible dead code. The `failure_reason` work in this PR doesn't depend on #292 for correctness — events still fire with sensible tags for every other branch.

## Diff size note: 
**~1,200 of the diff is manual-test scaffolding requested for this PR** — `AnalyticsTestBench.tsx` (~480 lines, one button per classifier branch), `analytics.test.ts` (~270 lines of classifier coverage), `analytics-echo-server.mjs` (~130 lines, the local echo server), and the playground README walkthrough (~90 lines). The actual *production* change is ~150 lines across `analytics.ts`, `requestRouter.ts`, and `multichain/index.ts`. Reviewers can skim past the test bench + echo server files — they're standalone, fully scoped to the playground, and don't ship to consumers.

## Summary

`mmconnect_wallet_action_failed` and `mmconnect_connection_failed` now carry a producer-side `failure_reason` tag describing **why** they fired. Today these events fire as a catch-all for anything that doesn't look like a user rejection — transport timeouts, WebSocket drops, wallet-side internal errors, malformed payloads, permission denials, etc. all land in the same bucket with no way to distinguish them in Mixpanel.

The tag is set via a new `classifyFailureReason(error)` helper. Initial taxonomy:

| Tag | Comes from |
|---|---|
| `transport_timeout` | `TransportTimeoutError` from `@metamask/multichain-api-client`; `new Error('Request timeout')` from `DefaultTransport`'s own setTimeout path. |
| `transport_disconnect` | `TransportError` from MWP; narrow message substring set (`'not connected'`, `'transport disconnect'`, `'connection lost'`, `'socket closed'`) so wallet messages mentioning "disconnect" don't leak in. |
| `wallet_unauthorized` | EIP-1193 `4100 Unauthorized` — the multichain permission layer ([core source](https://github.com/MetaMask/core/blob/main/packages/multichain-api-middleware/src/handlers/wallet-invokeMethod.ts)) rejects any method not in the granted CAIP-25 scope with this code, *before* the method handler runs. So on multichain this is the practical "method not allowed" signal — not `-32601`. Most dapp-side misuse (calling a method that wasn't in scope at connect time) will surface here once #292 lands and stops misclassifying `4100` as `_rejected`. |
| `wallet_method_unsupported` | Wallet JSON-RPC `-32601`, or EIP-1193 `4200 Unsupported Method`. Fires when the method *is* in scope but the wallet itself refuses or doesn't recognise it — less common than `wallet_unauthorized` on multichain sessions. |
| `wallet_invalid_params` | Wallet JSON-RPC `-32602`. Malformed payloads (e.g. non-hex address). |
| `wallet_internal_error` | Wallet JSON-RPC `-32603` (or anything in the `-32000 … -32099` server-error range). |
| `unknown` | Fallback for anything we can't classify — no recognised code, no recognised message. Includes EIP-1193 / EIP-1474 provider-defined codes we haven't carved out into their own bucket (e.g. `4900 Disconnected`); these can be promoted out of `unknown` later as the distribution justifies, without a schema migration. |

There's also an `unrecognised_chain` branch in the classifier (wallet `4902` or fallback message match) but it's effectively unreachable from a normal multichain dapp today — `wallet_switchEthereumChain` is rarely in the granted scope, so `4100` fires first. Left in the classifier as a no-op safety net.

The classifier checks the wallet JSON-RPC code **before** the transport-disconnect substring heuristic so a wallet error like `{ code: 4900, message: 'Disconnected' }` correctly lands in `unknown` rather than leaking into `transport_disconnect`. Pinned by a regression test.

A small `getUnwrappedErrorDetails` helper is introduced for the classifier so it sees wallet-side codes (e.g. `-32601`, `4200`) through the SDK's `RPCInvokeMethodErr` wrapper, which otherwise reports its static `code: 53`. `isRejectionError` itself is **not modified by this PR** — that work lives in #292.

Schema-side change (open string for now, no enum lock-in): [`metamask-sdk-analytics-api#31`](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/31). Once we have data on the distribution we can tighten this to a closed enum in a follow-up.

## Manual test bench

Added an **"Analytics test bench"** collapsible section at the top of the browser playground (`playground/browser-playground/src/components/AnalyticsTestBench.tsx`) with one button per classifier branch. Pair it with the new local echo server (`yarn analytics:echo`). Full walkthrough in the [playground README](../tree/wapi-1253-failure-reason-and-tracking-fixes/playground/browser-playground#manually-testing-analytics-events).

## Test plan

- `yarn test` passes on `@metamask/connect-multichain`, including the new `analytics.test.ts` cases for the classifier (every code branch + connection-side error fixtures + the `4900` / `transport_disconnect` ordering regression) and the new `requestRouter.test.ts` cases asserting `failure_reason` on transport timeouts and wallet internal errors.
- `yarn build`, `yarn lint:eslint`, and `yarn changelog:validate` all clean.
- Manually verified each classifier branch end-to-end in the browser playground via the new test bench + echo server.
- Reviewer: confirm we're OK with the open-string `failure_reason` schema for the first iteration (vs. a closed enum from day one).

## Schema mirror note

`packages/analytics/src/schema.ts` is auto-generated from `api.spec.yml` upstream, but was generated by an older `openapi-typescript` than what's installed in the workspace today (upstream emits `type` aliases; current version emits `interface`s). Regenerating produced a 1k-line diff of mechanical noise, so I hand-edited the file to add only the `failure_reason?: string` field. A future generator-version bump can do the full regen separately.

## Out of scope

- Modifying `isRejectionError` (unwrap `RPCInvokeMethodErr`, drop `4100`, tighten "user" heuristic) — owned by #292.
- Defining a closed enum for `failure_reason` — deliberately deferred so we can observe the real distribution first.
- **Tracking pre-flight failures** (`Transport not initialized` thrown from the `transport` getter when a dapp calls `invokeMethod` before connecting). These throw outside `RequestRouter`'s analytics wrapper so no `_failed` event fires today.
- **Tracking read-only RPC node failures.** `handleWithRpcNode` in `RequestRouter` throws without calling `analytics.track` today, so `RPCHttpErr` / `RPCReadonlyRequestErr` / `RPCReadonlyResponseErr` are not surfaced anywhere. Earlier iterations of this PR shipped dead `rpc_node_*` classifier branches; those were removed to keep `FailureReason` honest. When a producer wires up `handleWithRpcNode` tracking we will add the buckets back at the same time.
- Tracking pre-flight EVM errors (`IGNORED_METHODS` rejections inside `#requestInterceptor`) and `SDK_HANDLED_METHODS` errors. Documented in the WAPI-1253 spike doc; can fix in a follow-up.
- Adding an `error_code` / `error_message_sample` companion property on `_failed` events to make the `unknown` bucket triageable. Tracked as a separate follow-up PR (schema + producer).


packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -9,7 +9,58 @@ import type {
...
+  }
+
+  if (error instanceof RPCInvokeMethodErr) {
+    return {
+      code: error.rpcCode ?? error.code,
+      message: error.rpcMessage ?? error.message ?? '',
+    };
+  }
+
+  const errorObj = error as { code?: number; message?: string };
```
wenfix, line 143: [nit]: is there an existing type for this? If not, perhaps its overkill to create one

packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -36,6 +87,113 @@ export function isRejectionError(error: unknown): boolean {
...
+    }
+    if (code === 4200) {
+      // Unsupported method — wallet handler exists but explicitly refuses.
+      return 'wallet_method_unsupported';
+    }
+    if (code === 4902) {
+      // Unrecognized chain ID — `wallet_switchEthereumChain` to a chain the
+      // wallet hasn't been told about. Same signal as the message heuristic
+      // below, but reaches us cleanly via code rather than substring.
+      return 'unrecognised_chain';
```
wenfix: [nit]: maybe an enum for these?

playground/browser-playground/public/index.html
```diff
@@ -14,7 +14,7 @@
                script-src 'self' 'unsafe-inline' 'unsafe-eval';
                style-src 'self' 'unsafe-inline';
                img-src 'self' data:;
-               connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://mm-sdk-analytics.api.cx.metamask.io https://*.infura.io;
+               connect-src 'self' wss://mm-sdk-relay.api.cx.metamask.io https://mm-sdk-analytics.api.cx.metamask.io https://*.infura.io http://localhost:* http://127.0.0.1:*;
```
wenfix, line 26: [nit]: perhaps we should comment these to explain what they are?

wenfix (dismissed):
Left some non-blocking comments but otherwise LGTM

packages/connect-multichain/src/multichain/utils/analytics.ts
ffmcgee725: Seems to be intentional as per the docs on `FailureReason` in the previous file

```
 * Intentionally a string union (not a const enum) so callers stay free to
 * pass through a new bucket; the schema-side property is an open string for
 * the same reason.
```

playground/browser-playground/public/index.html
ffmcgee725: ➕

packages/connect-multichain/src/multichain/utils/analytics.ts
adonesky1: checked — `JsonRpcError` from `@metamask/utils` has required `code`/`message`, so `unknown` input would still need a cast. Same inline shape already shows up at `requestRouter.ts:147`, matches existing style. Leaving as-is.

packages/connect-multichain/src/multichain/utils/analytics.ts
adonesky1: yep this intentional for now

playground/browser-playground/public/index.html
adonesky1: Good call: https://github.com/MetaMask/connect-monorepo/pull/290/changes/525fde02dacef196c4f61fc447b5b8fbec0c9431

wenfix (dismissed):

packages/analytics/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add optional `failure_reason` property to `MMConnectProperties` in `schema.ts`, attached by producers on `mmconnect_wallet_action_failed` and `mmconnect_connection_failed`. Mirrors [`metamask-sdk-analytics-api#31`](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/31). ([#290](https://github.com/MetaMask/connect-monorepo/pull/290))
+- Add optional `error_code` (integer) and `error_message_sample` (string, max 200 chars) properties to `MMConnectProperties` in `schema.ts`. Producers attach them on the two `*_failed` events as raw-code / sanitised-message companions to `failure_reason`. Mirrors [`metamask-sdk-analytics-api#32`](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/32). ([#293](https://github.com/MetaMask/connect-monorepo/pull/293))
```
jiexi: technically part of #290 now, but nbd

packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -11,6 +11,111 @@ import type {
...
+ * the same reason.
+ */
+export type FailureReason =
+  | 'transport_timeout'
+  | 'transport_disconnect'
+  | 'wallet_method_unsupported'
+  | 'wallet_invalid_params'
+  | 'wallet_internal_error'
+  | 'wallet_unauthorized'
+  | 'unrecognised_chain'
```
jiexi: metamask is returning the american english error message, not the british english
```suggestion
  | 'unrecognized_chain'
```

packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -11,6 +11,111 @@ import type {
...
+ * suffix unscrubbed. Decimal-number scrubbing runs last so it doesn't
+ * fragment hex / Base58 tokens that contain digit runs.
+ */
+const SANITISE_PATTERNS: { pattern: RegExp; replacement: string }[] = [
+  // EVM-style 20-byte hex addresses (e.g. `0x` + 40 hex chars).
+  { pattern: /0x[a-fA-F0-9]{40}/gu, replacement: '<addr>' },
+  // Other long hex blobs: tx hashes, signatures, raw byte strings, large
+  // hex amounts. 16+ hex chars catches 32-byte hashes/signatures without
+  // snagging EVM method selectors (8 chars) or short hex codes.
+  { pattern: /(?:0x)?[a-fA-F0-9]{16,}/gu, replacement: '<hex>' },
```
jiexi, line 63: probably fine, but this would also replace words that are 16 characters or longer. I can't imagine what word that might be though

packages/connect-multichain/src/multichain/utils/analytics.ts
```diff
@@ -81,6 +186,150 @@ export function isRejectionError(error: unknown): boolean {
...
+    errorMessage.includes('not connected') ||
+    errorMessage.includes('transport disconnect') ||
+    errorMessage.includes('connection lost') ||
+    errorMessage.includes('socket closed')
+  ) {
+    return 'transport_disconnect';
+  }
+
+  if (errorMessage.includes('unrecognized chain')) {
+    return 'unrecognised_chain';
```
jiexi: does the `code === 4902` case not handle this already?

packages/connect-multichain/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Attach `error_code` and `error_message_sample` companion properties to `mmconnect_wallet_action_failed` and `mmconnect_connection_failed`. `error_code` preserves the raw wallet-side JSON-RPC / EIP-1193 code (e.g. `4001`, `-32603`) so the `unknown` `failure_reason` bucket stays triageable without a schema change every time a new code appears. `error_message_sample` is a sanitised, 200-char-max preview of the original error message — wallet addresses (EVM `0x{40}`, Solana / Bitcoin Base58 ≥32 chars, Bech32 with short HRPs like `bc1…` / `cosmos1…`), long hex blobs, URLs, and 10+-digit decimal numbers are scrubbed before emission. Both fields are optional and only set on the two `*_failed` events. The new `extractErrorDiagnostics` helper bundles all three diagnostic properties so producer call sites stay a single spread. Schema-side: [`metamask-sdk-analytics-api#32`](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/32). ([#293](https://github.com/MetaMask/connect-monorepo/pull/293))
```
jiexi: not sure if it's worth putting in so much detail about what exactly is being scrubbed. idk

packages/connect-multichain/src/multichain/utils/analytics.ts
adonesky1: good call — applied in d04d7b6. extension and mobile both emit the american spelling 😅

packages/connect-multichain/src/multichain/utils/analytics.ts
adonesky1: A 16+ char "word" of just abcdef + digits — basically can't happen in real wallet errors. upside (catching 32-byte hashes/sigs cleanly) outweighs the theoretical downside (a <hex> substitution in a sample where some sufficiently weird word existed — still triageable). leaving as-is; can tighten if we ever see one.

packages/connect-multichain/src/multichain/utils/analytics.ts
adonesky1: yeah, dead in real traffic. extension and mobile both always set code: 4902 on this error.  dropped the branch + the test in d04d7b6.

packages/analytics/CHANGELOG.md
adonesky1: good call — flipped to (#290) in c18807a. did the same on the parallel connect-multichain/CHANGELOG.md:13 entry for consistency.

jiexi (approved):

PR #291, opened 2026-05-11
refactor(connect-evm): stop double-tracking EVM intercepted methods

adonesky1:
## Summary

`MetamaskConnectEVM` was tracking `wallet_switchEthereumChain`, `wallet_addEthereumChain`, and `eth_requestAccounts` / `wallet_requestPermissions` at the EVM shim layer **in addition to** the multichain `RequestRouter`. Mixpanel was seeing **two** `_failed` events for one logical failure (and similarly two `_succeeded` / two `_requested`).

This PR removes the shim's tracking; the router is the single source of truth for these methods. The `@metamask/analytics` import and the `getWalletActionAnalyticsProperties` / `isRejectionError` helpers are no longer needed in `connect-evm`.

## Behaviour change to flag on review

On the `switchChain` → "Unrecognized chain ID" → `addEthereumChain` recovery path, the failed switch attempt **no longer fires a `_failed` event on its own**. The shim swallows the original `Unrecognized chain ID` error and recovers via `addEthereumChain`; only the successful add is tracked.

This matches the user's experience of one logical chain switch and was an explicit product call. If we want to keep visibility on switch-attempts-that-needed-recovery, we can add that as a separate dedicated event in a follow-up rather than re-introducing the duplicate `_failed`.

## Test plan

- [x] `yarn test` passes on `@metamask/connect-evm` (63 tests; no shim tracking assertions existed, so nothing to update).
- [x] `yarn build` succeeds across the workspace.
- [x] `yarn lint:eslint` and `yarn changelog:validate` clean.
- [ ] Reviewer: confirm we're OK with the recovery-path behaviour change.

## Related

- Spike: [WAPI-1253](https://consensyssoftware.atlassian.net/browse/WAPI-1253) — _Unpack all code paths that trigger `mmconnect_wallet_action_failed`_.
- Sibling PRs from the same spike (independent — can land in any order):
  - **This PR** — dedupe EVM shim tracking.
  - **Coming**: fix `isRejectionError` to unwrap `RPCInvokeMethodErr` and tighten the "user" heuristic.
  - **Coming**: attach `failure_reason` to `mmconnect_wallet_action_failed` / `mmconnect_connection_failed`.

## Out of scope

- `failure_reason` plumbing, classifier helper, schema changes — separate PR.
- `isRejectionError` unwrap fix — separate PR.

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Remove `@metamask/chain-agnostic-permission` dependency. The two helpers used from it (`getEthAccounts`, `getPermittedEthChainIds`) and the `parseScopeString` utility are now implemented locally on top of `@metamask/utils` primitives. This drops the transitive `@metamask/controller-utils` / `lodash` / `bn.js` / `eth-ens-namehash` / `fast-deep-equal` / `@metamask/ethjs-unit` chain from the `connect-evm` bundle.
+- Remove duplicate `mmconnect_wallet_action_*` tracking from `MetamaskConnectEVM.switchChain`, the `wallet_requestPermissions` / `eth_requestAccounts` interceptor branch, and `#addEthereumChain`. These methods are already tracked by the multichain `RequestRouter` when the request lands at the wallet, so the EVM shim was double-firing on Mixpanel. The `@metamask/analytics` package and the `getWalletActionAnalyticsProperties` / `isRejectionError` helpers are no longer imported here.
 
 ## [1.2.0]
 
--- a/packages/connect-evm/src/connect.ts
+++ b/packages/connect-evm/src/connect.ts
@@ -1,6 +1,5 @@
 /* eslint-disable no-restricted-syntax -- Private class properties use established patterns */
 /* eslint-disable @typescript-eslint/naming-convention -- __PACKAGE_VERSION__ is an esbuild define convention */
-import { analytics } from '@metamask/analytics';
 import type {
   MultichainCore,
   MultichainOptions,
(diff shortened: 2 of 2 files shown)


adonesky1:
Closing — the central premise of this PR is wrong on source review.

**The claim:** "EVM intercepted methods are tracked twice — once in the shim, once in the multichain `RequestRouter` when the call hits `wallet_invokeMethod`."

**The reality:** the EVM shim's intercepted-method paths (`switchChain`, `addEthereumChain`, `eth_requestAccounts` / `wallet_requestPermissions`) never reach `RequestRouter.invokeMethod`. They route through `this.#request(...)` → `transport.sendEip1193Message(...)`, which posts the raw EIP-1193 payload directly (`window.postMessage` on the default transport, `dappClient.sendRequest({ name: 'metamask-provider', ... })` on MWP) — no `wallet_invokeMethod` wrapping, no router invocation, no `_withAnalyticsTracking`. See:

- `packages/connect-evm/src/connect.ts:745-758` — `#request` calls `transport.sendEip1193Message`.
- `packages/connect-multichain/src/multichain/transports/default/index.ts:159-203` — raw postMessage.
- `packages/connect-multichain/src/multichain/transports/mwp/index.ts:388-427` — raw dappClient.sendRequest.
- `packages/connect-evm/src/provider.ts:73-75` — `INTERCEPTABLE_METHODS` are diverted to `#requestInterceptor`; only non-interceptable methods reach `core.invokeMethod` and the router.

So this PR isn't removing duplication — it's deleting the only `mmconnect_wallet_action_*` telemetry these four methods have. After merge, Mixpanel would see zero `_requested`/`_succeeded`/`_failed`/`_rejected` for `wallet_switchEthereumChain` and `wallet_addEthereumChain`, and only `mmconnect_connection_*` (different event family, different property shape) for `eth_requestAccounts` / `wallet_requestPermissions`.

The WAPI-1253 spike doc made this same claim, which is where the PR description came from — I'll retract it in the spike doc.

The **one** real duplication I found is the `switchChain → "Unrecognized chain ID" → addEthereumChain` recovery path, where the failed switch fires once *and* the recovery `addEthereumChain` fires once. Different bug. Tracked as a follow-up in `pr-tracker.md` / the spike doc; not addressing it here.

PR #295, opened 2026-05-12
feat: Reduce QR code density. Send QR Flow initialMessage as separate request

jiexi:
## Explanation

We're experiencing an issue where users are having difficulty scanning the QR code on their phone camera app. It seems Android is primarily affected, but the issue is somewhat noticeable on iOS as well. This is caused by the QR code becoming very dense due to the large wallet_createSession request in the sessionRequest.initialMessage property that contributed to the large size of the resulting deeplink URI. 

To fix this, desktop QR flows will no longer include the wallet_createSession request in the deeplink URI. Instead the desktop dapp will send the wallet_createSession request separately after it has received the signal that the MWP handshake has been completed on the wallet side. 

This change only works on the QR flow because we are guaranteed that the desktop webpage is still running and actively listening to the websocket messages. This is not true in mobile native browsers where the webpage may "go to sleep" when the user navigates away while following the deeplink to the wallet. Because of this, the native mobile browser deeplink (non QR MWP) flow must continue to include wallet_createSession in the sessionRequest.initialMessage property. Additionally the deeplink in the native browser flow does not have the same size limitation as this uri is opened directly, not encoded and decoded from a QR code, and so it doesn't really require this change in the first place

### Before QR code
Screenshot 2026-05-13 at 11 29 24 AM

### After QR Code
Screenshot 2026-05-13 at 11 27 43 AM

### Chrome Desktop QR flow
Screenshot 2026-05-12 at 12 00 11 PM

### Safari Mobile Native Deeplink flow
Screenshot 2026-05-12 at 11 58 02 AM

## Checklist

### DIFF
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Attach a `failure_reason` tag to `mmconnect_wallet_action_failed` and `mmconnect_connection_failed` events via a new `classifyFailureReason` helper, distinguishing transport timeouts, transport disconnects, EIP-1193 wallet errors (`4100 wallet_unauthorized`, `4200 wallet_method_unsupported`, `4902 unrecognized_chain`), and JSON-RPC wallet errors (`-32601`, `-32602`, `-32603`, plus the `-32000…-32099` server-error range), with an `unknown` fallback. Schema-side: [`metamask-sdk-analytics-api#31`](https://github.com/consensys-vertical-apps/metamask-sdk-analytics-api/pull/31). ([#290](https://github.com/MetaMask/connect-monorepo/pull/290))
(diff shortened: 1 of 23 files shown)


packages/connect-multichain/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Improves QR code scanning reliability. Desktop QR code MWP flow now omits the iniital request from the deeplink URI, instead sending it as a separate request after the wallet completes the MWP handshake. This results in a shorter deeplink URI and less dense QR code. Native deeplink (non-QR MWP) flow is unchanged. ([#295](https://github.com/MetaMask/connect-monorepo/pull/295))
```
adonesky1:
```suggestion
- Improves QR code scanning reliability. Desktop QR code MWP flow now omits the initial request from the deeplink URI, instead sending it as a separate request after the wallet completes the MWP handshake. This results in a shorter deeplink URI and less dense QR code. Native deeplink (non-QR MWP) flow is unchanged. ([#295](https://github.com/MetaMask/connect-monorepo/pull/295))
```

adonesky1 (dismissed):
This looks great. Only thing: maybe we should add a unit test or two for this?

wenfix (approved):

PR #303, opened 2026-05-20
feat: allow disabling analytics

wenfix:
## Explanation

Adds an `analytics.enabled` option across the public connect packages so consumers can fully disable dapp-side analytics.

Analytics remain enabled by default. When `analytics.enabled` is `false`, the SDK disables the analytics singleton, clears queued sender events, skips analytics global property setup and event tracking, and omits the MWP `analytics.remote_session_id` connection metadata.

`@metamask/connect-evm` and `@metamask/connect-solana` now forward the option to `@metamask/connect-multichain`. The package READMEs and root CSP guidance now document the option and clarify that the analytics endpoint is only needed when analytics are enabled.

## References

Closes [WAPI-1504](https://consensyssoftware.atlassian.net/browse/WAPI-1504).

## Checklist

- [ ] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/connect-monorepo/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add an `analytics.enabled` option to `createEVMClient()`. Set it to `false` to disable dapp-side analytics events and wallet correlation metadata. ([#303](https://github.com/MetaMask/connect-monorepo/pull/303))
+- Add an `analytics.enabled` option to `createEVMClient()`. Consumers can set it to `false` to disable dapp-side analytics events and wallet correlation metadata. ([#303](https://github.com/MetaMask/connect-monorepo/pull/303))
 
 ## [1.3.1]
 
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add an `analytics.enabled` option to `createEVMClient()`. Set it to `false` to disable dapp-side analytics events and wallet correlation metadata. ([#303](https://github.com/MetaMask/connect-monorepo/pull/303))
+
 ## [1.3.1]
 
 ### Fixed
--- a/packages/connect-evm/README.md
+++ b/packages/connect-evm/README.md
@@ -200,9 +200,6 @@ const client = await createEVMClient({
     accountsChanged: (accounts) => console.log('Accounts:', accounts),
     chainChanged: (chainId) => console.log('Chain:', chainId),
   },
-  analytics: {
-    enabled: false,
-  },
   debug: true,
 });
 ```
--- a/packages/connect-evm/README.md
+++ b/packages/connect-evm/README.md
@@ -165,21 +165,23 @@ Factory function to create a new MetaMask Connect EVM instance.
 
 #### Parameters
 
-| Option                     | Type                                          | Required | Description                                          |
-| -------------------------- | --------------------------------------------- | -------- | ---------------------------------------------------- |
-| `dapp.name`                | `string`                                      | Yes      | Name of your dApp                                    |
-| `api.supportedNetworks`    | `Record<Hex, string>`                         | Yes      | Map of hex chain IDs to RPC URLs                     |
-| `dapp.url`                 | `string`                                      | No       | URL of your dApp                                     |
-| `dapp.iconUrl`             | `string`                                      | No       | Icon URL for your dApp                               |
-| `ui.headless`              | `boolean`                                     | No       | Run without UI (for custom QR implementations)       |
-| `ui.preferExtension`       | `boolean`                                     | No       | Prefer browser extension over mobile (default: true) |
-| `ui.showInstallModal`      | `boolean`                                     | No       | Show installation modal for desktop                  |
-| `mobile.preferredOpenLink` | `(deeplink: string, target?: string) => void` | No       | Custom deeplink handler                              |
-| `mobile.useDeeplink`       | `boolean`                                     | No       | Use `metamask://` instead of universal links         |
-| `transport.extensionId`    | `string`                                      | No       | Custom extension ID                                  |
-| `transport.onNotification` | `(notification: unknown) => void`             | No       | Notification handler                                 |
-| `eventHandlers`            | `Partial<EventHandlers>`                      | No       | Event handlers for provider events                   |
-| `debug`                    | `boolean`                                     | No       | Enable debug logging                                 |
+| Option                      | Type                                          | Required | Description                                                                                                                  |
+| --------------------------- | --------------------------------------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `dapp.name`                 | `string`                                      | Yes      | Name of your dApp                                                                                                            |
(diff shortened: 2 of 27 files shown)


packages/connect-multichain/src/multichain/index.ts
```diff
@@ -222,7 +237,20 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     return instancePromise;
   }
 
+  /**
+   * Sets up analytics globals for the current SDK options.
+   */
+  public async setupAnalytics(): Promise<void> {
+    await this.#setupAnalytics();
+  }
```
jiexi: can we just make the `#setupAnalytics` no longer hash tag private 😅

packages/connect-multichain/README.md
```diff
@@ -171,6 +172,9 @@ const client = await createMultichainClient({
       'eip155:137': 'https://polygon-mainnet.infura.io/v3/KEY',
     },
   },
+  analytics: {
+    enabled: false,
+  },
```
jiexi:
```suggestion
```

packages/connect-solana/README.md
```diff
@@ -102,19 +102,33 @@ Creates a new Solana client instance. By default, the wallet is automatically re
...
 
+```typescript
+const client = await createSolanaClient({
+  dapp: {
+    name: 'My Solana DApp',
+    url: 'https://mydapp.com',
+  },
+  analytics: {
+    enabled: false,
+  },
```
jiexi:
```suggestion
```

packages/connect-evm/README.md
```diff
@@ -198,6 +200,9 @@ const client = await createEVMClient({
     accountsChanged: (accounts) => console.log('Accounts:', accounts),
     chainChanged: (chainId) => console.log('Chain:', chainId),
   },
+  analytics: {
+    enabled: false,
+  },
```
jiexi:
```suggestion
```

jiexi:
is the intention that if the existing MultichainClient singleton gets param merged `analytics.enabled` to be false that analytics would be disabled at that point by calling `disable()` on the Analytics class? If so, i'm not sure i'm seeing where in code that's facilitated.

I imagine we want to make it so that analytics can only be flipped to off, but never back to on via merge options?

wenfix:
thanks @jiexi, I've addressed the feedback in ec897c32302717d08be6b596647cc076533509fe & 7fce56696280b4ded051a5664c5e1ef7f5559bf5

packages/connect-evm/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add an `analytics.enabled` option to `createEVMClient()`. Set it to `false` to disable dapp-side analytics events and wallet correlation metadata. ([#303](https://github.com/MetaMask/connect-monorepo/pull/303))
```
adonesky1:
```suggestion
- Add an `analytics.enabled` option to `createEVMClient()`. Consumers can set it to `false` to disable dapp-side analytics events and wallet correlation metadata. ([#303](https://github.com/MetaMask/connect-monorepo/pull/303))
```

packages/connect-evm/src/connect.ts
```diff
@@ -1008,6 +1008,7 @@ export class MetamaskConnectEVM {
  * @param options.dapp - Dapp identification and branding settings
```
adonesky1, line 1030: It seems like in the cases of shimmed methods like `wallet_switchEthereumChain` and `wallet_addEthereumChain` we're still calling `this.#trackWalletActionRequested`, ` #trackWalletActionSucceeded` etc. Look at `#requestInterceptor`

adonesky1: I guess the analytics themselves are not emitted since they will be gated in the `track` method on the analytics class, but they will still call into `getWalletActionAnalyticsProperties` and pull from storage and we're creating an unnecessary divergence in patterns.

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -222,7 +237,16 @@ export class MetaMaskConnectMultichain extends MultichainCore {
     return instancePromise;
   }
 
-  async #setupAnalytics(): Promise<void> {
+  /**
+   * Sets up analytics globals for the current SDK options.
+   */
+  public async setupAnalytics(): Promise<void> {
```
adonesky1: why did we make this public?

adonesky1: Just for testing?

packages/connect-evm/src/connect.ts
adonesky1: I think the pattern should be we just gate at the top of all of the `#trackWalletAction*` methods like we did `track` on the analytics class and then don't add logic to avoid calling them?

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -184,15 +207,7 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
-      analytics.setGlobalProperty(
-        'mmconnect_versions',
-        instance.options.versions ?? {},
-      );
-      if (options.analytics?.integrationType) {
-        analytics.setGlobalProperty('integration_types', [
-          options.analytics.integrationType,
-        ]);
-      }
+      await instance.setupAnalytics();
```
adonesky1: This can crash in the duplicate-package case that this global singleton is meant to handle. If an older bundled copy of `@metamask/connect-multichain` creates `globalThis.__METAMASK_CONNECT_MULTICHAIN_SINGLETON__` first, this newer copy will reuse that older instance and call `instance.setupAnalytics()`. But released older instances don’t have that new public method, so it throws `TypeError: instance.setupAnalytics is not a function`.

Admittedly this is pretty unlikely, but worth protecting against. Could we keep this backwards compatible by not making the existing-singleton branch depend on a new method on the reused instance? That instance may come from an older/different bundled copy, so this branch should stick to APIs/fields that older singletons already had, or otherwise safely skip/fallback so older singleton instances don’t crash.

packages/connect-evm/src/connect.ts
wenfix: Fixed in MetaMask/connect-monorepo@bcac2d6ec44b823ba784f040573a1e309da94447 by gating each EVM `#trackWalletAction*` method before building wallet-action analytics properties.

packages/connect-evm/CHANGELOG.md
wenfix: Fixed in MetaMask/connect-monorepo@04f16b0fe86cadb0d8beb4f413b5eea18b618a54.

packages/connect-multichain/src/multichain/index.ts
wenfix: Fixed in MetaMask/connect-monorepo@e4f2160f79babbc9ab9f226714ea847fb100d5b3 by guarding the reused singleton analytics hook and adding a regression test for older singleton instances without `setupAnalytics`.

packages/connect-multichain/src/multichain/index.ts
wenfix: Fixed in MetaMask/connect-monorepo@e4f2160f79babbc9ab9f226714ea847fb100d5b3. `setupAnalytics` is now TypeScript-private and only used as an internal singleton compatibility hook.

jiexi (approved):

adonesky1 (approved):

PR #304, opened 2026-05-21
feat(connect-evm): auto-announce EIP-6963 provider

wenfix:
## Explanation

`@metamask/connect-evm` did not participate in EIP-6963 wallet discovery after the SDK loaded. In raw EIP-6963 consumers, that could produce duplicate MetaMask entries when the native MetaMask provider was already present, or no MMConnect-managed MetaMask entry when the extension was absent.

This PR adds an EIP-6963 announcer for the MMConnect-managed EIP-1193 provider:

- Announces the MMConnect-managed provider as `MetaMask` with `rdns: "io.metamask.mmc"` when no native MetaMask provider is observed.
- Suppresses MMConnect-managed announcement when `io.metamask` or `io.metamask.mobile` is observed.
- Adds `skipAutoAnnounce` and `announceProvider()` for manual control.
- Re-announces the same provider detail on later `eip6963:requestProvider` events.
- Uses a package-local MetaMask fox SVG asset bundled as a data URI for EIP-6963 metadata. (we could potentially create a shared asset package instead of duplicating)

The implementation is browser-only and scoped to the current window/frame. It does not mutate `window.ethereum`.

## Review Updates

- Added an EIP-6963 validation panel to the browser playground for observing announcements, dispatching `eip6963:requestProvider`, manually calling `legacySDK.announceProvider()`, and confirming whether the observed provider is the legacy EVM SDK provider.
- Updated the browser playground wagmi configuration to pass `skipAutoAnnounce: true`, so the validation dapp does not produce duplicate MMConnect-managed announcements from both the direct legacy EVM SDK path and the wagmi connector path.
- Updated `connect-multichain` extension detection to exact-match native MetaMask EIP-6963 RDNS values (`io.metamask`, `io.metamask.mobile`) instead of using `startsWith("io.metamask")`, so `io.metamask.mmc` cannot accidentally select the browser-extension/default transport.

## Release / Product Note

This is API-additive, but it intentionally changes wallet discovery behavior for dapps that consume EIP-6963. Release/product owners should confirm whether this should be classified as a visible behavior change or breaking change before merge.

## References

- Closes [WAPI-1501](https://consensyssoftware.atlassian.net/browse/WAPI-1501)
- EIP-6963 spec: https://eips.ethereum.org/EIPS/eip-6963

## Checklist

- [x] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes (N/A: no consumer PRs required unless release/product owners reclassify this visible behavior change)


packages/connect-evm/src/eip6963.ts
```diff
@@ -0,0 +1,355 @@
...
+
+/**
+ * Display name used for the SDK-managed MetaMask EIP-6963 provider.
+ */
+export const CONNECT_EVM_EIP6963_NAME = 'MetaMask';
+
+/**
+ * Reverse-DNS identifier used for the SDK-managed MetaMask EIP-6963 provider.
+ */
+export const CONNECT_EVM_EIP6963_RDNS = 'io.metamask.mmc';
```
adonesky1, line 29: Where is this string coming from?

adonesky1: Ok you just came up with it fresh

packages/connect-evm/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Announce the SDK-managed EIP-1193 provider through EIP-6963 by default when native MetaMask has not already announced, with `skipAutoAnnounce` and `announceProvider()` for manual control. ([#304](https://github.com/MetaMask/connect-monorepo/pull/304))
```
adonesky1: [nit] ?
```suggestion
- Announce the MMConnect-managed EIP-1193 provider through EIP-6963 by default when native MetaMask has not already announced, with `skipAutoAnnounce` and `announceProvider()` for manual control. ([#304](https://github.com/MetaMask/connect-monorepo/pull/304))
```

packages/connect-evm/README.md
```diff
@@ -306,6 +310,22 @@ None.
 await client.disconnect();
 ```
 
+##### `announceProvider()`
+
+Announces the SDK-managed EIP-1193 provider through EIP-6963 unless a native MetaMask provider has already announced with `rdns` `io.metamask` or `io.metamask.mobile`. This is called automatically by `createEVMClient()` unless `skipAutoAnnounce: true` is set. The first call may take up to 300 ms while native providers are requested.
```
adonesky1:
```suggestion
Announces the MMConnect-managed EIP-1193 provider through EIP-6963 unless a native MetaMask provider has already announced with `rdns` `io.metamask` or `io.metamask.mobile`. This is called automatically by `createEVMClient()` unless `skipAutoAnnounce: true` is set. The first call may take up to 300 ms while native providers are requested.
```

adonesky1:
Should we add a flow to test this in our the browser playground?

packages/connect-evm/CHANGELOG.md
adonesky1: Maybe I'm overthinking it but should this be considered breaking? We're now defaulting to announcing a new event that might change dapp side UX?

packages/connect-evm/README.md
```diff
@@ -46,6 +46,9 @@ const client = await createEVMClient({
   },
 });
 
+// The SDK-managed provider is announced through EIP-6963 by default.
```
adonesky1:
```suggestion
// The MMConnect-managed provider is announced through EIP-6963 by default.
```

packages/connect-evm/src/eip6963.ts
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  return [...announcedRdns].some(isNativeMetaMaskRdns);
+};
+
+/**
+ * Announces an SDK-managed EIP-1193 provider through EIP-6963.
+ *
+ * Announcement is best-effort and browser-only. Native MetaMask providers with
+ * `io.metamask` or `io.metamask.mobile` suppress this SDK provider to avoid
```
adonesky1, line 235:
```suggestion
 * `io.metamask` or `io.metamask.mobile` suppress this announcement of this provider to avoid
```

packages/connect-evm/src/eip6963.ts
```diff
@@ -0,0 +1,355 @@
...
+  byte.toString(16).padStart(2, '0'),
+);
+
+/**
+ * Formats random bytes as an RFC 4122 version 4 UUID.
+ *
+ * @param bytes - Random bytes from a cryptographically secure source.
+ * @returns A UUID string.
+ */
+const formatUuidV4 = (bytes: Uint8Array): string => {
```
adonesky1, line 107: 🤔 this utility seems a bit strange to inline, but I'm guessing this was an effort to not pull in additional hefty packages unnecessarily?

packages/connect-evm/src/eip6963.ts
adonesky1: This rdns matches [our is extension installed check](https://github.com/MetaMask/connect-monorepo/blob/54135ce8cdcd9e78dfe92012a4a472100e02f6fb/packages/connect-multichain/src/domain/platform/index.ts#L121-L123).

If somehow the provider is already announced, could we end up using the default transport by mistake? Maybe we should use a non-colliding rdns?

jiexi: I think RDNS needs to match either `io.metamask` or `io.metamask.mobile` because otherwise Wagmi will see this as another connector / won't dedupe it

packages/connect-evm/src/eip6963.ts
```diff
@@ -0,0 +1,355 @@
...
+ * Reverse-DNS identifier used for the SDK-managed MetaMask EIP-6963 provider.
+ */
+export const CONNECT_EVM_EIP6963_RDNS = 'io.metamask.mmc';
+
+/**
+ * Native MetaMask EIP-6963 identifiers that should suppress SDK announcement.
+ */
+export const METAMASK_EIP6963_RDNS = [
+  'io.metamask',
+  'io.metamask.mobile',
```
jiexi, line 36: technically there are other variants here, but it's probably best we only consider these two cases for now

packages/connect-evm/README.md
wenfix: Applied this wording in the README.

packages/connect-evm/CHANGELOG.md
wenfix: Applied the `MMConnect-managed` wording in the changelog. I also updated the PR body with a Release / Product Note: this is API-additive, but it intentionally changes EIP-6963 wallet discovery UX, so release/product owners should confirm classification before merge.

packages/connect-evm/README.md
wenfix: Applied this wording in the README.

packages/connect-evm/src/eip6963.ts
wenfix: Yes. The helper is intentionally local to avoid adding a UUID dependency for this browser-only best-effort path. It uses Web Crypto (`randomUUID` or `getRandomValues`) and throws if Web Crypto is unavailable, which avoids the CodeQL issue and avoids falling back to weak randomness.

packages/connect-evm/src/eip6963.ts
wenfix: Leaving the current wording slightly different: `Native MetaMask providers with io.metamask or io.metamask.mobile suppress this SDK provider...` avoids the duplicate phrase in the suggested text while preserving the intent.

wenfix:
Added a browser playground EIP-6963 test bench and README validation steps. It observes `eip6963:announceProvider`, can dispatch `eip6963:requestProvider`, can manually call `legacySDK.announceProvider()`, and marks whether the announcement provider matches the legacy EVM provider. Also set playground wagmi `skipAutoAnnounce: true` to avoid direct SDK plus wagmi duplicate announcements during validation.

adonesky1:
Screenshot 2026-05-28 at 12 48 17 PM

What is `legacy provider` mean in this context? I see it as `yes` for mmconnect but nothing else?

packages/connect-multichain/src/ui/ModalFactory.ts
```diff
@@ -176,7 +176,9 @@ export abstract class BaseModalFactory<
         this.displayUriCallback?.(newLink);
         return newLink;
       },
-      onClose: this.onCloseModal.bind(this),
+      onClose: (shouldTerminate) => {
+        this.onCloseModal(shouldTerminate).catch(() => undefined);
+      },
```
adonesky1: can you explain why this change was necessary and/or how it relates to the other 6963 changes?

packages/connect-multichain/src/ui/index.ts
```diff
@@ -17,7 +17,7 @@ export async function preload(): Promise<void> {
     const { defineCustomElements } = await import(
       '@metamask/multichain-ui/loader'
     );
-    await defineCustomElements();
+    defineCustomElements();
```
adonesky1, line 20: can you explain why this change was necessary and/or how it relates to the other 6963 changes?

adonesky1:
Reminder to also set `skipAutoAnnounce: true` in the wagmi connector at `integrations/wagmi/metamask-connector.ts`  and I've created a ticket to make the update on the wagmi side once this PR is merged: https://consensyssoftware.atlassian.net/browse/WAPI-1532

packages/connect-multichain/src/domain/platform/index.ts
```diff
@@ -105,21 +110,24 @@ const detectionPromise: Promise<boolean> = (async () => {
 
   return new Promise((resolve) => {
     const providers: any[] = [];
+    const targetWindow = window;
```
adonesky1, line 114: [nit] worth a one-line comment on what this guards against — the new test ("uses the captured window if the global window is removed before detection completes") implies it's protecting against the global being torn down between request and timeout (SPA navigation, vitest unstubAllGlobals, etc.)

playground/browser-playground/src/components/Eip6963TestBench.tsx
```diff
@@ -0,0 +1,243 @@
+import { useEffect, useRef, useState } from 'react';
+import { TEST_IDS } from '@metamask/playground-ui';
+import type {
+  EIP1193Provider,
+  MetamaskConnectEVM,
+} from '@metamask/connect-evm';
+
+const EIP6963_ANNOUNCE_PROVIDER_EVENT = 'eip6963:announceProvider';
+const EIP6963_REQUEST_PROVIDER_EVENT = 'eip6963:requestProvider';
```
adonesky1: [nit] Should we just pull these from `@metamask/connect-evm` ?

packages/connect-multichain/src/ui/index.ts
wenfix: Removed this unrelated preload change in 896ef60a. It was not necessary for the EIP-6963 behavior.

packages/connect-multichain/src/ui/ModalFactory.ts
wenfix: Removed this unrelated UI change in 896ef60a. It was not needed for the EIP-6963 work.

packages/connect-multichain/src/domain/platform/index.ts
wenfix: Added the one-line comment in 896ef60a to explain that the captured window guards against the global being torn down before the detection timeout completes.

playground/browser-playground/src/components/Eip6963TestBench.tsx
wenfix: Updated in 83554ef9: the EIP-6963 event constants are now exported from @metamask/connect-evm and the browser playground imports them from the package instead of duplicating the strings.

wenfix:
Follow-up on the latest top-level comments: renamed the playground column from `Legacy provider` to `MMConnect SDK provider` in 83554ef9, and set `skipAutoAnnounce: true` by default in `integrations/wagmi/metamask-connector.ts` in e863ae27 while preserving an explicit override. Added connector-level test coverage for that wagmi default.

packages/connect-evm/src/eip6963.ts
adonesky1: Been going back and forth on this... I feel like its possible it might be useful for this provider to have its own distingushable rdns in the future, but there isn't actually a case I can think of right now. We will resolve the deduping issue by setting `skipAutoAnnounce` to true in the WAGMI connector with https://consensyssoftware.atlassian.net/browse/WAPI-1532... but yeah I don't know. I don't think it matters too much so if you feel strongly on this @jiexi I'm happy to run with `io.metamask` here

jiexi: i think this is probably fine by skipping the autoAnnounce in the Wagmi connector. We can always adjust later

jiexi (approved):

PR #305, opened 2026-05-21
refactor: cleanup deeply nested event callback promises

jiexi:
## Explanation

Refactors several offenders of the `new Promise()` and event callback soup pattern. This PR replaces that pattern with a deferred promise to reduce the amount of nesting and by breaking down `MWPTransport.connect()` into smaller methods. 

Extension and Mobile E2E tests are passing.

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/index.ts
+++ b/packages/connect-multichain/src/multichain/index.ts
@@ -19,6 +19,8 @@ import {
   METAMASK_DEEPLINK_BASE,
   MWP_RELAY_URL,
 } from '../config';
+import { createDeferredPromise } from '@metamask/utils';
+
 import {
   getVersion,
   type InvokeMethodOptions,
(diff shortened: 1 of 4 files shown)


packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -1,8 +1,7 @@
-/* eslint-disable @typescript-eslint/prefer-promise-reject-errors */
 /* eslint-disable consistent-return */
-/* eslint-disable promise/param-names */
+
 /* eslint-disable @typescript-eslint/no-misused-promises */
-/* eslint-disable @typescript-eslint/no-shadow */
+
```
adonesky1: [total nit] don't need these blank lines probably

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -304,82 +303,245 @@ export class MWPTransport implements ExtendedTransport {
...
+      } catch (err) {
+        resumeDeferred.reject(err);
+      }
+    };
+
+    if (this.dappClient.state === 'CONNECTED') {
+      runOnResumeHandler();
+    } else {
+      this.dappClient.once('connected', runOnResumeHandler);
+      this.dappClient.resume(session.id ?? '');
```
adonesky1:
```suggestion
      this.dappClient.resume(session.id ?? '').catch((err) => resumeDeferred.reject(err));
```

`resume()` can reject ([per `BaseClient.resume` types: `SessionError` when the session is missing/expired or the client isn't `DISCONNECTED`](https://github.com/MetaMask/mobile-wallet-protocol/blob/%40metamask%2Fmobile-wallet-protocol-core%400.4.0/packages/core/src/base-client.ts#L86-L106)). without this catch the error is swallowed, `'connected'` never fires, and the caller waits the full `resumeTimeout` only to see a generic `TransportTimeoutError` instead of the real cause.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -304,82 +303,245 @@ export class MWPTransport implements ExtendedTransport {
...
-      } else if (!walletSession) {
-        // TODO: verify if this branching logic can ever be hit
+  private async onResumeHandler(options?: {
+    scopes: Scope[];
+    caipAccountIds: CaipAccountId[];
+    forceRequest?: boolean;
+  }): Promise<void> {
+    await this.waitForWalletSessionIfNotCached();
+    const sessionResponse = await this.request({ method: 'wallet_getSession' });
+    // TODO: verify if this branching logic can ever be hit
```
adonesky1: do we still know whether this todo is relevant 😬 😅 ?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -304,82 +303,245 @@ export class MWPTransport implements ExtendedTransport {
...
+        // Issue of this is it will send a session update event with an empty session and right after we may get the session recovered
+        // await this.request({ method: 'wallet_revokeSession', params: walletSession });
         walletSession = response.result as SessionData;
       }
-      await this.removeStoredPendingSessionRequest();
-      this.notifyCallbacks({
-        method: 'wallet_sessionChanged',
-        params: walletSession,
+    } else if (!walletSession) {
+      // TODO: verify if this branching logic can ever be hit
```
adonesky1: do we still know whether this todo is relevant 😬 😅 ?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -304,82 +303,245 @@ export class MWPTransport implements ExtendedTransport {
...
+    const timeoutDeferred = createDeferredPromise<never>();
+    const timeout = setTimeout(
+      () => timeoutDeferred.reject(new TransportTimeoutError()),
+      this.options.resumeTimeout,
+    );
+
+    return Promise.race([
+      resumeDeferred.promise,
+      timeoutDeferred.promise,
+    ]).finally(() => clearTimeout(timeout));
```
adonesky1:
```suggestion
   const cleanup = () => this.dappClient.off('connected', runOnResumeHandler);

    return Promise.race([
      resumeDeferred.promise,
      timeoutDeferred.promise,
    ]).finally(() => { clearTimeout(timeout); cleanup(); } );
```

if the timeout wins the race we still leave a `'connected'` listener registered. it'll fire on the next reconnect and re-run `onResumeHandler` — sending `wallet_getSession` / `wallet_createSession` for a connect attempt the caller already gave up on. removing it in `.finally` keeps the listener scoped to this attempt.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -304,82 +303,245 @@ export class MWPTransport implements ExtendedTransport {
...
-            return resumeReject(new Error(response.error.message));
-          }
-          // TODO: Maybe find a better way to revoke sessions on wallet without triggering an empty notification
-          // Issue of this is it will send a session update event with an empty session and right after we may get the session recovered
-          // await this.request({ method: 'wallet_revokeSession', params: walletSession });
-          walletSession = response.result as SessionData;
-        }
-      } else if (!walletSession) {
-        // TODO: verify if this branching logic can ever be hit
+  private async onResumeHandler(options?: {
```
adonesky1: MIght as well make it match the pattern for the rest of the file while we're in here?
```suggestion
  async #onResumeHandler(options?: {
```

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/305/commits/094fe2d377358fc2d17443b2ddb259467c2aa64a

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: still relevant. I don't think it's possible for wallet_getSession to error, even when returning from cache. But not 100% sure

jiexi: keeping this for safety

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: yes, relevant, but highly unlikely.

https://github.com/MetaMask/connect-monorepo/pull/305/commits/1af081a2c0dd8c20a61e17868a42bdcc098e4dba

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -336,82 +333,256 @@ export class MWPTransport implements ExtendedTransport {
...
       }
-      await this.removeStoredPendingSessionRequest();
-      this.notifyCallbacks({
-        method: 'wallet_sessionChanged',
-        params: walletSession,
+    } else if (!walletSession) {
+      // Hitting this branch implies that the MWP session was established,
+      // but the user has not yet accepted the initial wallet_createSession approval,
+      // but the page has refreshed and we've lost that previous context and so we
+      // are trying to recover by making a new wallet_createSession request.
```
adonesky1, line 383: 🙇 I appreciate you adding this context

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -336,82 +333,256 @@ export class MWPTransport implements ExtendedTransport {
...
+      this.dappClient
+        .resume(session.id ?? '')
+        .catch((err) => resumeDeferred.reject(err));
     }
+
+    const timeoutDeferred = createDeferredPromise<never>();
+    const timeout = setTimeout(
+      () => timeoutDeferred.reject(new TransportTimeoutError()),
+      this.options.resumeTimeout,
+    );
```
adonesky1, line 450: I think this quietly changes the resume-path timeout, despite the "No behavior change" note.

on `main`, `connect()` picked the timeout once for both paths:

```ts
storedPendingSessionRequest ? this.options.resumeTimeout : this.options.connectionTimeout
```

the split moved that logic into the two helpers, but only `#startSession` kept the conditional. `#resumeSession` hardcodes `resumeTimeout`:

```ts
const timeout = setTimeout(
  () => timeoutDeferred.reject(new TransportTimeoutError()),
  this.options.resumeTimeout, // was: storedPending ? resumeTimeout : connectionTimeout
);
```

`getActiveSession()` (the resume trigger) and the pending-request key are independent, so a normal resume has no pending request → it used to fall back to `connectionTimeout` (120s) and now always gets `resumeTimeout` (10s).

happy-path resume resolves instantly so most users won't notice. but a resume can hit `wallet_createSession` — `forceRequest`, the `!walletSession` recovery branch, or (as in the video) requesting different scopes/accounts than the current session. that needs a human to approve in the wallet, and they now have 10s instead of 120s before it times out.

fix is either restore the conditional in `#resumeSession`, or if 10s-on-resume is intentional, drop the "No behavior change" line and call it out.

repro (video): connect + approve, then re-`connect()` with a different scope/account set, and wait ~15s before approving → times out at ~10s on this branch; same flow on `main` waits.

https://github.com/user-attachments/assets/96c2c880-432e-4ada-bb18-ca060171825d

packages/connect-multichain/src/multichain/index.ts
```diff
@@ -608,57 +614,44 @@ export class MetaMaskConnectMultichain extends MultichainCore {
...
-      // Listen for session_request to generate and emit the QR code link
-      this.dappClient.on(
-        'session_request',
-        (sessionRequest: SessionRequest) => {
-          const connectionRequest: ConnectionRequest = {
-            sessionRequest,
-            metadata: this.#buildConnectionMetadata(),
-          };
+    // Listen for session_request to generate and emit the QR code link
+    this.dappClient.on('session_request', (sessionRequest: SessionRequest) => {
```
adonesky1, line 625: Pre-existing, but maybe we want to handle it now? We never `.off` this guy?

packages/connect-multichain/src/multichain/transports/mwp/index.ts
jiexi: https://github.com/MetaMask/connect-monorepo/pull/305/commits/b4daea60c0e8d17bd9a2c90cc5209d891cc53333

packages/connect-multichain/src/multichain/index.ts
jiexi: doing this in a separate PR

jiexi: https://github.com/MetaMask/connect-monorepo/pull/314

adonesky1 (approved):
LGTM! Awesome work here @jiexi as always 🙇 🚀

PR #307, opened 2026-05-26
refactor(connect-multichain): rename storage transport type methods

wenfix:
https://consensyssoftware.atlassian.net/browse/WAPI-1527

## Summary
- Rename the Store API for persisted transport values from get/set/removeTransport to get/set/removeTransportType.
- Update MetaMaskConnectMultichain call sites and related tests to use the clearer transport type method names.
- Preserve the existing multichain-transport storage key, so no persisted data migration is needed.
- Add the connect-multichain changelog entry required for the exported API rename.

## Acceptance Notes
- The renamed API describes the stored value as a transport type string instead of implying a transport instance.
- All production call sites and tests now use getTransportType, setTransportType, and removeTransportType.
- A search for the old method calls only finds the backlog context file, which is untracked and not part of this PR.

## Test Plan
- [x] yarn workspace @metamask/connect-multichain build
- [x] yarn test:unit (from packages/connect-multichain)
- [x] yarn workspace @metamask/connect-multichain changelog:validate

### DIFF
--- a/packages/connect-multichain/CHANGELOG.md
+++ b/packages/connect-multichain/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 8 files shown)


jiexi (dismissed):

adonesky1 (approved):
LGTM

PR #310, opened 2026-05-27
fix(connect-evm): normalize intercepted return shapes

wenfix:
## Summary
- Return EIP-2255-style requested permissions from intercepted `wallet_requestPermissions` provider calls.
- Return null from successful `wallet_switchEthereumChain` provider calls.
- Return null from successful `wallet_addEthereumChain` provider calls.

## Testing
- yarn workspace @metamask/connect-evm test:unit
- yarn tsc --noEmit -p packages/connect-evm/tsconfig.json


packages/connect-evm/src/connect.ts
```diff
@@ -676,6 +697,9 @@ export class MetamaskConnectEVM {
         if (request.method === 'eth_requestAccounts') {
           return result.accounts;
         }
+        if (request.method === 'wallet_requestPermissions') {
+          return getRequestedPermissions(params);
```
jiexi: this isn't going to return a properly faked wallet_requestPermissions result unfortunately. We need to fake both `eth_accounts` and `permittedChains` here

```
[
    {
        "id": "jbUYFMYDfFkNd6EYIyA1h",
        "parentCapability": "eth_accounts",
        "invoker": "https://metamask.github.io",
        "caveats": [
            {
                "type": "restrictReturnedAccounts",
                "value": [
                    "0xee166a3eec4796dec6a1d314e7485a52bbe68e4d"
                ]
            }
        ],
        "date": 1779916966390
    },
    {
        "id": "jbUYFMYDfFkNd6EYIyA1h",
        "parentCapability": "endowment:permitted-chains",
        "invoker": "https://metamask.github.io",
        "caveats": [
            {
                "type": "restrictNetworkSwitching",
                "value": [
                    "0x1",
                    "0x10e6",
                    "0x2105",
                    "0x38",
                    "0xa",
                    "0xa4b1",
                    "0xe708"
                ]
            }
        ],
        "date": 1779916966390
    }
]
```

packages/connect-multichain/src/domain/platform/index.test.ts
jiexi, line 1: looks like some changes from the other PR got mixed into this one

packages/connect-multichain/src/domain/platform/index.ts
jiexi, line 1: looks like some changes from the other PR got mixed into this one

packages/connect-multichain/CHANGELOG.md
jiexi, line 1: looks like some changes from the other PR got mixed into this one

wenfix: 26dd70e

packages/connect-multichain/src/domain/platform/index.ts
wenfix: 26dd70e

packages/connect-multichain/src/domain/platform/index.test.ts
wenfix: 26dd70e

packages/connect-evm/src/connect.ts
wenfix: db570e9b521791f1f8b0f98b03985199f4ccf991

packages/connect-evm/src/connect.ts
```diff
@@ -68,6 +68,90 @@ type ConnectOptions = {
...
+    return globalThis.crypto.randomUUID();
+  }
+
+  return `${Date.now().toString(36)}-${Math.random().toString(36).slice(2)}`;
+};
+
+const getDappInvoker = (options: MultichainOptions): string => {
+  const dappUrl = options?.dapp?.url;
+  if (!dappUrl) {
+    return '';
```
adonesky1: `dapp.url` is optional and ReactNative Apps use `nativeScheme` (or just `name`) instead, so this returns '' for them and both will end up with `invoker: ''`. Even though this is just what we fake back to the app we might as well get it right and it should be the requesting origin. Can we fall back `url` origin → `nativeScheme` → `name`? worth a test for the no-url case too.

packages/connect-evm/src/connect.ts
```diff
@@ -68,6 +68,90 @@ type ConnectOptions = {
...
+  invoker: string;
+}): RequestedPermission[] => {
+  const permissionRequest = Array.isArray(params) ? params[0] : undefined;
+  if (
+    !permissionRequest ||
+    typeof permissionRequest !== 'object' ||
+    Array.isArray(permissionRequest) ||
+    !('eth_accounts' in permissionRequest)
+  ) {
+    return [];
```
adonesky1: [minor/non-blocking] since `isConnectRequest` matches by method name, a `wallet_requestPermissions` call without `eth_accounts` (e.g. a `permitted-chains`-only request, or malformed params) still forces a connect prompt and then returns `[]` here — so the user approves but the dapp gets empty `[]`. realistically almost nobody calls it that way against this SDK since we couple accounts+chains at connect, so fine to leave as-is, but just highlighting that this is possible. If we want to have something defensive simplest is to not force the prompt when we'd just return `[]` in `#requestInterceptor`: 

```
if (isConnectRequest(request)) {
  // bail before prompting if this requestPermissions call can't produce a result
  if (
    request.method === 'wallet_requestPermissions' &&
    !requestsEthAccounts(request.params)
  ) {
    return []; // or throw unsupported-permission — but no connect() call, so no prompt
  }

  const shouldForceConnectionRequest =
    request.method === 'wallet_requestPermissions';
  // ... existing chainIds setup ...
  const result = await this.connect({ ... }); // <- the prompt; now never reached for the empty case
```

adonesky1: downside is it make the interceptor a bit more convoluted than it already is

adonesky1 (commented):
One small blocking comment/issue re providing acccurate "invoker" value for non browser based apps, otherwise LGTM

packages/connect-evm/src/connect.ts
jiexi: IMO we shouldn't bother validating the request param shape nor honoring the values requested. Too much complexity. As long as we return the response/results the right shape I think that's sufficient

wenfix: Leaving this unchanged for this PR. The SDK connect flow couples accounts and chains, and adding request-shape validation would make the interceptor branchier; keeping the scope to returning the correct response shape.

packages/connect-evm/src/connect.ts
wenfix: Fixed in e5f0b835: getDappInvoker now falls back url origin -> nativeScheme -> name, with tests for the no-url nativeScheme and name cases.

packages/connect-evm/src/connect.ts
wenfix: This is addressed in the current diff: wallet_requestPermissions now fakes both eth_accounts and endowment:permitted-chains, and connect.test.ts covers both entries.

adonesky1 (dismissed):
LGTM!

packages/connect-evm/src/connect.ts
jiexi: The first element in `permissionRequest` needs to have an object with key `eth_accounts`. `permissionRequest` isn't an array of strings.

```
await window.ethereum.request({method:'wallet_requestPermissions', params: [{'eth_accounts': {}}]})
```

Again. I think we shouldn't validate the param shape or value at all

adonesky1: done in b0d53f1 — ripped out the param validation. connect always grants accounts+chains anyway, so checking the request shape didn't change the output, just added branches. nice side effect: no more `[]`-after-prompt case. added a test for a non-`eth_accounts` request to lock it in.

jiexi (approved):

PR #311, opened 2026-05-27
fix: reject MWP response errors

wenfix:
## Explanation

DefaultTransport rejects wallet errors, while MWPTransport could resolve response-shaped wallet errors nested under result.error. That forced ecosystem clients to handle both thrown and returned error forms.

This normalizes MWP pending request handling so response-shaped wallet errors reject with the parsed wallet error, preserving numeric wallet error codes. It also removes downstream returned-error checks from RequestRouter and connect-evm switchChain so callers consume one transport contract: success resolves, failure rejects.

## References

- WAPI-1520

## Checklist

- [x] I've prepared draft pull requests for clients and consumer packages to resolve any breaking changes (not applicable: no consumer breaking changes)

## Test Plan

- [x] yarn workspace @metamask/connect-multichain test:unit
- [x] yarn workspace @metamask/connect-multichain build
- [x] yarn workspace @metamask/connect-evm test:unit
- [x] yarn workspace @metamask/connect-evm build
- [x] yarn workspace @metamask/connect-multichain changelog:validate
- [x] yarn workspace @metamask/connect-evm changelog:validate

### DIFF
--- a/packages/connect-multichain/src/multichain/rpc/requestRouter.ts
+++ b/packages/connect-multichain/src/multichain/rpc/requestRouter.ts
@@ -118,15 +118,6 @@ export class RequestRouter {
       }
 
       const response = await request;
-      if (response.error) {
-        const { error } = response;
-        throw new RPCInvokeMethodErr(
-          `RPC Request failed with code ${error.code}: ${error.message}`,
-          error.code,
-          error.message,
-        );
-      }
-
       return response.result as Json;
     });
   }
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.ts
@@ -539,10 +539,12 @@ export class MWPTransport implements ExtendedTransport {
                 return;
               }
 
+              const responseError = this.getResponseError(messagePayload);
+
               // Handle error response (e.g., user rejected the connection)
-              if (messagePayload.error) {
+              if (responseError) {
                 return rejectConnection(
-                  this.parseWalletError(messagePayload.error),
+                  this.parseWalletError(responseError),
                 );
               }
 
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.ts
@@ -210,6 +210,37 @@ export class MWPTransport implements ExtendedTransport {
     return rpcErrors.internal({ message });
   }
 
+  private getResponseError(messagePayload: Record<string, unknown>): unknown {
+    if ('error' in messagePayload && messagePayload.error) {
+      return messagePayload.error;
+    }
+
+    const { result } = messagePayload;
+    if (
+      typeof result === 'object' &&
+      result !== null &&
+      'error' in result &&
+      result.error &&
+      this.isErrorPayload(result.error)
+    ) {
+      return result.error;
+    }
+
+    return undefined;
+  }
+
+  private isErrorPayload(errorPayload: unknown): boolean {
+    if (errorPayload instanceof Error) {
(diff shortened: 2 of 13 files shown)


packages/connect-multichain/src/multichain/rpc/requestRouter.ts
```diff
@@ -118,15 +118,6 @@ export class RequestRouter {
...
       const response = await request;
-      if (response.error) {
-        const { error } = response;
-        throw new RPCInvokeMethodErr(
-          `RPC Request failed with code ${error.code}: ${error.message}`,
-          error.code,
-          error.message,
-        );
-      }
-
```
adonesky1, line 129: Not sure I follow this removal?

adonesky1: ah ok the idea is that the transport rejects before we get here

adonesky1: ok so after some back and forth with machine god:

Removing this guard assumes every `ExtendedTransport.request()` rejects wallet errors, but `DefaultTransport.request()` still delegates directly to `@metamask/multichain-api-client`, whose transports resolve raw `{ error }` responses. That means extension/default-transport wallet failures could now fall through as `response.result === undefined` and be tracked as successes. We should introduce the same reject-on-wallet-error symmetry in `DefaultTransport.request()` before removing this router-level fallback.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -221,10 +252,11 @@ export class MWPTransport implements ExtendedTransport {
...
             clearTimeout(request.timeout);
 
-            // Check if the message contains an error (e.g., user rejected)
-            if ('error' in messagePayload && messagePayload.error) {
+            const responseError = this.getResponseError(messagePayload);
+
+            if (responseError) {
               this.pendingRequests.delete(messagePayload.id);
-              request.reject(this.parseWalletError(messagePayload.error));
+              request.reject(this.parseWalletError(responseError));
```
adonesky1, line 259: There's another instance below (lines 542-547 on this file in this PR) in `initialConnectionMessageHandler` where we should be passing the return from `this.getResponseError(messagePayload)` rather than  `messagePayload.error` directly

adonesky1: and can we also add a test on the QR + inline connect paths where wallet_createSession returns `{ result: { error: ... } }`? right now that nested shape is only covered through `handleMessage`, not `initialConnectionMessageHandler`

wenfix: Fixed in cfcf2972. `initialConnectionMessageHandler` now passes `this.getResponseError(messagePayload)` into `parseWalletError`, so both top-level `error` and nested `result.error` shapes are rejected consistently. Added regression coverage for both QR-code and inline initial `wallet_createSession` paths.

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
wenfix: Fixed in cfcf2972. `DefaultTransport.request()` now rejects top-level wallet `{ error }` responses from the inner multichain-api-client transport and preserves the numeric code, so the router-level fallback can stay removed. Added regression coverage in `src/multichain/transports/default/index.test.ts`.

adonesky1 (approved):
Ok I've tested this against the test dapp a bunch on both MWP and Default transports and its working as expected and code looks good. Last thing I'm wondering is have you validated whether this breaks any of our e2e test?

PR #312, opened 2026-05-28
chore: normalize wallet invocation errors across transports

wenfix:
## Explanation

`connect-multichain.invokeMethod()` could surface different error shapes depending on whether a wallet error arrived as a resolved JSON-RPC error response or as a rejected transport error. That also meant `connect-evm` could lose provider-facing details when adapting multichain errors to EIP-1193 request errors.

This change normalizes wallet invocation failures at the `RequestRouter` boundary:

- extracts wallet `code`, `message`, and JSON-RPC `data` from response errors and wrapped transport errors
- walks a capped `cause` chain so transport/API wrappers do not hide the wallet error
- wraps normalized wallet failures in `RPCInvokeMethodErr` while preserving `reason`, `rpcCode`, `rpcMessage`, and `rpcData`
- keeps existing `RPCInvokeMethodErr` instances unchanged
- keeps analytics classification using normalized wallet codes while preserving original uncoded transport errors for transport diagnostics

`connect-evm` now maps normalized `RPCInvokeMethodErr` instances into EIP-1193-facing errors that carry the wallet `code`, provider-facing `message`, and JSON-RPC `data` when present.

## References

- WAPI-1521

## Checklist


packages/connect-evm/src/provider.test.ts
```diff
@@ -21,23 +21,23 @@ function createMockCore(
...
-    it('re-throws RPCInvokeMethodErr as an EIP-1193 error with the correct code', async () => {
+    it('maps normalized RPCInvokeMethodErr to an EIP-1193 error with the wallet code', async () => {
       const mockCore = createMockCore();
       const provider = new EIP1193Provider(mockCore as any, vi.fn());
       provider.selectedChainId = '0x1';
 
       mockCore.invokeMethod.mockRejectedValue(
         new RPCInvokeMethodErr(
-          'RPC Request failed with code 4001: User denied transaction signature.',
+          'User rejected the request',
```
jiexi: `User rejected the request.`

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
```diff
@@ -30,22 +31,161 @@ import type { RpcClient } from './handlers/rpcClient';
...
+    const codedDetails = getCodedErrorDetails(currentObject);
+    if (codedDetails) {
+      const descendantObjects = errorObjectChain.slice(index + 1);
+      const ancestorObjects = errorObjectChain.slice(0, index);
+      const descendantMessage = getFirstNonEmptyMessage(
+        descendantObjects.map((object) => object.message),
+      );
+      const ancestorMessage = getFirstNonEmptyMessage([
+        primitiveMessage,
+        ...ancestorObjects.map((object) => object.message),
```
jiexi: i'm not certain it makes sense to try to resolve an error message from earlier up in the chain. Hard for me to fully understand this problem though. What were your thoughts around this?

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
```diff
@@ -30,22 +31,161 @@ import type { RpcClient } from './handlers/rpcClient';
...
+}
+
+/**
+ * Extracts the public invocation error fields from either a JSON-RPC error
+ * payload or a rejected Error-like value.
+ *
+ * @param error - Unknown error thrown or returned during method execution.
+ * @returns Canonical fields for RPCInvokeMethodErr.
+ */
+function getInvocationErrorDetails(error: unknown): InvocationErrorDetails {
```
jiexi: thoughts on moving all these newly added error helpers into their own file?

packages/connect-multichain/src/ui/index.ts
jiexi, line 1: seems unrelated to this main purpose of this PR

packages/connect-multichain/src/ui/ModalFactory.ts
jiexi, line 1: seems unrelated to this main purpose of this PR

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
jiexi: quite a bit of complexity in this helper. How deeply nested are we expecting causes to be?

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
wenfix: Good instinct, you're right. The rpc fields were sourced from anywhere in the cause chain, so a transport/API wrapper message could leak into `rpcMessage` and pose as the wallet's protocol message. Fixed in 1527a1e1: `rpcCode`/`rpcMessage`/`rpcData` now read only from the coded wallet node; `reason` still falls back through the chain for readable logs. When the wallet sends no message, `rpcMessage` is left unset.

packages/connect-multichain/src/multichain/rpc/requestRouter.ts
wenfix: Agreed — moved them into `invocationError.ts` (exporting `toRPCInvokeMethodErr`) in b4b34482, so the router stays focused and the normalization logic is testable in isolation.

wenfix: Shallow in practice (wallet → transport → maybe an API wrapper, ~2–3 levels); the depth-5 cap is just a guard against cyclic chains, not an expected depth. Most of the complexity was the message juggling feeding `rpcMessage` — that's gone now that the rpc fields read only from the coded node (1527a1e1), so the remaining walk is just "find the first node with a numeric `code`."

packages/connect-multichain/src/ui/index.ts
wenfix: Agreed — this is no longer part of the PR. It turned out the branch already matched `main` for this file (main made the same change independently), so there's nothing left here after the merge.

packages/connect-evm/src/provider.test.ts
wenfix: Good catch — aligned the fixture to the canonical `'User rejected the request.'` (with the trailing period) in 685619ed.

adonesky1:
The `error.data` plumbing looks like it can't actually deliver `data` on either transport.

Claude traced it end to end. The wallet *does* send it: `wallet_invokeMethod` returns a top-level JSON-RPC `error` and the extension/mobile `serializeError` keeps `data`, so the transport receives `{ error: { code, message, data } }`. But both SDK transports drop it before normalization:

- `DefaultTransport.request` throws `#parseWalletError(response.error)` → `new Error(message)` + `code`, no `data`/`cause` (`default/index.ts`).
- `MWPTransport` rejects via `parseWalletError` → `new JsonRpcError(code, message)`, no `data` (`mwp/index.ts`).

Since both **reject** (rather than resolve an error envelope), the new `if (response.error)` guard in `handleWithWallet` never fires for them — so `toRPCInvokeMethodErr` only ever sees the stripped error, `rpcData` stays `undefined`, and `connect-evm` never sets `error.data`. The new router test passes only because it mocks the transport to *resolve* `{ error: { …, data } }`, which neither real transport does.

If we want the changelog claim to hold, the fix would need to live in the transports' `parseWalletError` (pass `data` as the 3rd `JsonRpcError` arg, or set `cause`) so the bytes survive to the router. 


**EDIT:** pushed what I + claude think we need here: https://github.com/MetaMask/connect-monorepo/pull/312/commits/432d04c7fddde13d8792d32e04ff6404be883913

jiexi (approved):

PR #314, opened 2026-05-28
fix(connect-multichain): remove session_request listener after #headlessConnect settles

jiexi:
## Explanation

Address review feedback from #305 in a separate PR.

The `session_request` listener registered in `MetaMaskConnectMultichain.#headlessConnect()` is added via `dappClient.on()` but never removed. Each subsequent `connect()` call (e.g. retry, account switch, or scope change) layers on another copy that will re-emit `display_uri` with stale deeplinks for every future `session_request` until the `dappClient` is GC'd.

Capture the handler as a named ref and `.off()` it in a `.finally()` so the listener is torn down on success, error, and timeout.

## References

- Review comment: https://github.com/MetaMask/connect-monorepo/pull/305#discussion_r3322023025

## Checklist

  - No new tests — this is a leak fix; existing tests still pass.
- [ ] I've communicated my changes to consumers by updating changelogs for packages I've changed, highlighting breaking changes as necessary
  - Will add changelog entry once PR number is known.

Made with [Cursor](https://cursor.com)


adonesky1 (approved):
LGTM!

PR #318, opened 2026-06-02
refactor: Remove MultichainClient.transport and transport.onNotification constructor param

jiexi:
This PR tightens the public surface of `@metamask/connect-multichain` so that all wallet RPC traffic flows through a single, documented entry point — `client.invokeMethod({ scope, request })` — and ecosystem clients (`@metamask/connect-evm`) no longer reach into transport internals.
### What changes (in dependency order)
1. **Remove `transport.onNotification` constructor option** from `createMultichainClient()` and `createEVMClient()`. The option was a fan-out of every typed event already exposed via `client.on(...)` (`stateChanged`, `wallet_sessionChanged`, `metamask_accountsChanged`, `metamask_chainChanged`, `display_uri`) — it was redundant and only kept consumers tied to a transport-shaped notification format. Migration: subscribe via `client.on('<event>', handler)` (multichain) or the EIP-1193 provider events / `eventHandlers` (EVM). The `transport.onNotification(callback)` method on the live transport instance is unaffected.
2. **Decouple `MultichainApiClientWrapperTransport` from `MetaMaskConnectMultichain`'s private state.** Previously the wrapper called `this.metamaskConnectMultichain.transport` directly, which became a problem the moment we wanted to remove the public getter. The wrapper now takes a `getTransport: () => ExtendedTransport | undefined` lambda in its constructor and re-resolves the live transport on every call.
3. **Route EIP-1193 / legacy provider methods through `invokeMethod`.** A new `EIP1193_PASSTHROUGH_METHODS` set (`wallet_addEthereumChain`, `wallet_switchEthereumChain`, `eth_accounts`) is recognized by `RequestRouter.invokeMethod` and dispatched to `transport.sendEip1193Message` directly, bypassing the `wallet_invokeMethod` envelope that the Multichain API would otherwise wrap them in. These three methods are wallet-side concerns the Multichain API does not model:
    - `wallet_addEthereumChain` / `wallet_switchEthereumChain` mutate the wallet's chain configuration, not session scope state.
    - `eth_accounts` is occasionally needed as a live read against the wallet (see EVM `#onSessionChanged`) outside the cached SDK session.
   This is intentionally placed in `RequestRouter` rather than `MetaMaskConnectMultichain.invokeMethod` so that any future consumer constructing a router has the same dispatch behavior, and so `invokeMethod` on the multichain class stays a one-line "build router → delegate".
   No multichain-side analytics are emitted for the passthrough path — `connect-evm` already emits its own `wallet_action_*` events around `switchChain` / `addEthereumChain`, and adding router-level tracking would double-count.
4. **Remove the public `transport` accessor** from `MultichainCore` and `MetaMaskConnectMultichain`. With (3) in place, the only documented use case for reaching into the transport from outside the SDK was to dispatch those three methods, and that now goes through `invokeMethod`. Internally the multichain class uses a private `#transportOrThrow()` helper.
5. **Migrate `connect-evm` off the public transport.** `MetamaskConnectEVM.#request` and `#onSessionChanged` now go through `this.#core.invokeMethod({ scope, request })` instead of `this.#core.transport.sendEip1193Message(...)`. `switchChain` and `#addEthereumChain` now thread `scope` through `#request` so the multichain analytics layer has scope context. Behavior, response shapes, error codes (including the `4902` "Unrecognized chain ID" fallback), and deeplink open behavior are all preserved.


packages/connect-multichain/src/multichain/rpc/requestRouter.ts
```diff
@@ -84,6 +88,30 @@ export class RequestRouter {
...
+   * @param options
+   */
+  private async handleWithEip1193Passthrough(
+    options: InvokeMethodOptions,
+  ): Promise<Json> {
+    const response = await this.transport.sendEip1193Message({
+      method: options.request.method,
+      params: options.request.params,
+    });
+    return response as Json;
```
adonesky1: `invokeMethod` now returns two different shapes depending on the method, which is a bit of a footgun on what this PR is making the public entry point.

`handleWithEip1193Passthrough` returns the full `{ id, jsonrpc, result }` envelope, but `handleWithWallet` (every other method) returns the unwrapped `result`.

Since the only consumer is internal (connect-evm), we should normalize the return rather than baking the asymmetry into the public API: unwrap in the passthrough (`return (response as { result: Json }).result`) and drop the `.result` in the three EVM call sites (`switchChain`, `#addEthereumChain`, `#onSessionChanged`). The `=== null` success checks still hold since the unwrapped value is `null`.

One optional extra you could drop in or leave for a follow-up: `eth_accounts` is now in both `SDK_HANDLED_METHODS` and `EIP1193_PASSTHROUGH_METHODS` (passthrough wins by ordering) — slightly confusing, but out of scope for this thread.

adonesky1: Branch off of this one with a demo on the playground showing how this plays out in effect:

`alex/WAPI-1516-invokemethod-return-shape-demo`

https://github.com/MetaMask/connect-monorepo/tree/alex/WAPI-1516-invokemethod-return-shape-demo

Screenshot 2026-06-05 at 1 37 49 PM

packages/connect-multichain/src/domain/multichain/api/constants.ts
```diff
@@ -168,3 +168,17 @@ export const RPC_HANDLED_METHODS = new Set([
...
+// `sendEip1193Message` (raw method/params, native extension/MWP postMessage routing).
+// These are wallet-side concerns that the Multichain API does not model:
+// - `wallet_addEthereumChain` / `wallet_switchEthereumChain` mutate the wallet's
+//   chain configuration, not session scope state.
+// - `eth_accounts` is occasionally needed as a live read against the wallet
+//   (see EVM `#onSessionChanged`) outside the cached SDK session state.
+export const EIP1193_PASSTHROUGH_METHODS = new Set([
+  'wallet_addEthereumChain',
+  'wallet_switchEthereumChain',
+  'eth_accounts',
```
adonesky1, line 183: **Not blocking**:  `eth_accounts` is now in both `SDK_HANDLED_METHODS` and `EIP1193_PASSTHROUGH_METHODS` — passthrough wins by ordering so it's harmless.

adonesky1 (approved):
LGTM! 🚀

PR #323, opened 2026-06-10
revert: Revert moving connect-multichain into peerDep for ecosystem packages

jiexi:
## Explanation

Reverts just the peerDep changes from 7786de8aeb349648033abbd3986652de01974710 . The devex degradation for library consumers having to install two peerDeps (ecosystem client, and multichain client) isn't worth the gains from build time warnings. 

## References

Fixes: https://consensys.slack.com/archives/C058P09HT17/p1781040721508819?thread_ts=1781037006.244619&cid=C058P09HT17

## Checklist

### DIFF
--- a/packages/connect-evm/CHANGELOG.md
+++ b/packages/connect-evm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `@metamask/connect-multichain` is no longer a peer dependency and will be installed transitively as a regular dependency ([#323](https://github.com/MetaMask/connect-monorepo/pull/323))
+
 ### Fixed
 
 - `EIP1193Provider.request()` now preserves JSON-RPC error `data` on provider-facing errors when `connect-multichain` surfaces a wallet error through `RPCInvokeMethodErr`. Dapps can read revert reasons, custom-error bytes, and other wallet-provided error metadata from `error.data` alongside the original wallet `error.code` and `error.message`. ([#312](https://github.com/MetaMask/connect-monorepo/pull/312))
--- a/packages/connect-evm/package.json
+++ b/packages/connect-evm/package.json
@@ -55,12 +55,12 @@
   },
   "dependencies": {
     "@metamask/analytics": "workspace:^",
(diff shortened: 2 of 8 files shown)


packages/connect-evm/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `@metamask/connect-multichain` is no longer a peer dependency and will be installed transitively as a regular dependency ([#323](https://github.com/MetaMask/connect-monorepo/pull/323))
```
adonesky1, line 12: Trying to reason about whether this is breaking again or not 🤔

adonesky1: Its not.

adonesky1 (approved):
LGTM!

wenfix:
Now that `connect-multichain` is a regular `workspace:^` dependency again, the build-time/runtime version-check feature from #253 is effectively inert — `__CONNECT_MULTICHAIN_PEER_VERSION_RANGE__` is still computed (now from `dependencies`) and checked at runtime, but the resolved range will basically always be satisfied by the installed transitive version, so the mismatch warning won't realistically fire anymore. The existing tests still pass since they stub the global directly.

The relevant spots:
- `packages/connect-evm/src/connect.ts:1225-1234`
- `packages/connect-solana/src/connect.ts:98-100`

Not a blocker for this revert — just flagging that the versioning machinery is now mostly dead weight. Worth a quick decision on whether to keep it for future re-introduction or strip it in a follow-up.

wenfix (approved):

wenfix:
Correction — you're right, I was wrong to call it inert. I missed that `connect-multichain` is a `globalThis`-keyed singleton (`packages/connect-multichain/src/multichain/index.ts:253-268`): `create()` returns the *existing* instance whenever one is already initialized, so the version actually handed back to the ecosystem client can differ from the one it was built against. That's exactly the mismatch the runtime check at `connect.ts:1230` is meant to catch.

If anything, moving back to a regular `dependency` makes that scenario *more* likely (duplicate copies in the tree now decide which version wins the singleton race), so the check earns its keep. Disregard my earlier suggestion to strip it.

PR #333, opened 2026-06-16
fix(connect-multichain): emit connection_rejected on QR modal close

wenfix:
## Explanation

When a user closes the MWP install / QR code modal during a connection attempt, the SDK tracked an `mmconnect_connection_failed` analytics event. Closing the modal is a user-driven cancellation, so it should be classified as `mmconnect_connection_rejected` instead. The bug polluted the `_failed` event (meant for genuine errors like transport timeouts/disconnects) with user cancellations, skewing the connection funnel and failure-reason metrics.

**Root cause:** `BaseModalFactory.onCloseModal` rejected the connection with `new Error('User closed modal')`. That error bubbles up to `#handleConnection`, which calls `isRejectionError(error)` to branch between `_rejected` and `_failed`. `isRejectionError` only matches code `4001` or the substrings `reject` / `denied` / `cancel` / `user rejected` / etc. — the message `"User closed modal"` matches none, so it fell through to `mmconnect_connection_failed`.

**Fix:** `onCloseModal` now rejects with `providerErrors.userRejectedRequest()` (the canonical EIP-1193 `4001` "User rejected the request" error from `@metamask/rpc-errors`, already a dependency of this package). This makes the cancellation intent explicit at the source rather than relying on string heuristics. The error flows directly through the success callback — no transport-boundary wrapping — so the `4001` code reaches `isRejectionError` intact, where it matches on both `code === 4001` and the `reject` substring. Genuine connection errors (transport timeout, transport disconnect, wallet RPC errors) are untouched and still emit `mmconnect_connection_failed`.

This is "Option 1" from the ticket, which was preferred over broadening the `isRejectionError` heuristics.

## References

- Fixes [WAPI-1552](https://consensyssoftware.atlassian.net/browse/WAPI-1552)

## Checklist

- [x] I've communicated my changes to consumers by [updating changelogs for packages I've changed](https://github.com/MetaMask/connect-monorepo/tree/main/docs/contributing.md#updating-changelogs), highlighting breaking changes as necessary

### DIFF
--- a/packages/connect-multichain/src/multichain/utils/analytics.test.ts
+++ b/packages/connect-multichain/src/multichain/utils/analytics.test.ts
@@ -1,4 +1,5 @@
 /* eslint-disable id-length -- vitest alias */
+import { providerErrors } from '@metamask/rpc-errors';
 import * as t from 'vitest';
 
 import {
@@ -485,6 +486,13 @@ t.describe('isRejectionError', () => {
       );
     },
   );
+
+  t.it('treats the modal-close error as a rejection', () => {
+    // BaseModalFactory.onCloseModal rejects with this when the user closes
+    // the QR/install modal mid-connection. It must classify as `_rejected`,
+    // not `_failed`.
+    t.expect(isRejectionError(providerErrors.userRejectedRequest())).toBe(true);
+  });
 
   t.it(
     'unwraps RPCInvokeMethodErr to detect wallet-side rejection codes',
--- a/packages/connect-multichain/src/multichain/utils/analytics.test.ts
+++ b/packages/connect-multichain/src/multichain/utils/analytics.test.ts
@@ -487,7 +487,7 @@ t.describe('isRejectionError', () => {
     },
   );
 
-  t.it('treats the modal-close error as a rejection', () => {
+  t.it('recognises userRejectedRequest (4001) as a rejection', () => {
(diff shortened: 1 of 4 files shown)


packages/connect-multichain/src/ui/index.test.ts
```diff
@@ -383,6 +387,12 @@ t.describe('ModalFactory', () => {
 
         // Multichain SDK is what will close the modal instead
         t.expect(mockModal.unmount).toHaveBeenCalled();
+
+        // Closing the modal is a user-driven cancellation: the error handed to
+        // the success callback must classify as a rejection so the SDK emits
+        // `mmconnect_connection_rejected`, not `_failed` (WAPI-1552).
```
adonesky1:
```suggestion
```

packages/connect-multichain/src/multichain/utils/analytics.test.ts
```diff
@@ -486,6 +487,13 @@ t.describe('isRejectionError', () => {
     },
   );
 
+  t.it('treats the modal-close error as a rejection', () => {
```
adonesky1:
```suggestion
  t.it('recognises userRejectedRequest (4001) as a rejection', () => {
```

adonesky1 (approved):
LGTM

PR #348, opened 2026-09-10
fix(connect-multichain): wipe MWP pairing on full disconnect (MCWP-822)

tommasini:
## Explanation

`MWPTransport.disconnect()`'s full-disconnect branch only called `dappClient.disconnect()`, relying on the wallet-protocol client's in-memory `session` to be non-null. `dappClient.disconnect()` is a no-op when that in-memory session is already `null` — e.g. after a prior failed `resume()`, which nulls the session without clearing `SessionStore`. In that case the pairing (relay channel + ECIES keys) survived in `SessionStore`.

The next `connect()` reads that leftover pairing via `getActiveSession()` (`SessionStore.list()[0]`), takes the resume path, and tries to resume a pairing the wallet has already revoked. That resume attempt times out, and the dapp is stuck — surviving even a page refresh, since the CAIP-25 session cache and the MWP pairing are two independent persisted stores. Only wiping storage manually recovered.

This reproduced for Fireblocks (`@metamask/connect-multichain@1.2.0`) on a successful connect → sign → `disconnect()` → `connect()`.

### Fix

- Added `MWPTransport.#clearMwpPairing()`: on a full disconnect (no remaining scopes), list and delete every `SessionStore` entry directly, independent of `dappClient`'s in-memory session state. `dappClient.disconnect()` is still called afterward for transport/channel cleanup, and its rejection is intentionally left to propagate — this preserves the existing behavior where a failed disconnect surfaces to callers (see `MWPTransport.disconnect()`'s existing `should handle disconnect errors` test).
- Full disconnect now also clears the persisted `pending_session_request`, matching what `connect()` already does in its `finally`, so a later `connect()` can't pick up a dead handshake URI.
- Defense in depth: if `#resumeSession` fails (e.g. the wallet already revoked the pairing out-of-band), that specific pairing is now deleted from `SessionStore` before rethrowing. This complements #180's existing reject-path cleanup on `#startSession`.
- Added unit test coverage for `disconnect()` in `mwp/index.test.ts` (previously untested): full disconnect wipes the pairing and `getActiveSession()` resolves `undefined` afterward, partial disconnect leaves the pairing untouched, and a subsequent `connect()` takes `#startSession` rather than resume.

Scope is intentionally limited to `@metamask/connect-multichain` in this repo — the leftover pairing lives in the dapp-side adapter (localStorage / IndexedDB / AsyncStorage), not in any Mobile wallet-side store, so `app/core/SDKConnectV2` in `metamask-mobile` is unaffected and out of scope here.

image

## References

* Fixes [MCWP-822](https://consensyssoftware.atlassian.net/browse/MCWP-822)
* Related to [MCWP-823](https://consensyssoftware.atlassian.net/browse/MCWP-823) (status=`connected` with `#transport` undefined) and [MCWP-821](https://consensyssoftware.atlassian.net/browse/MCWP-821) (stale IndexedDB handle) — same underlying report, not fixed by this change
* Complements [#180](https://github.com/MetaMask/connect-monorepo/pull/180) (cleans `SessionStore` on connect rejection); this PR closes the same gap on the disconnect path
* Epic: [MCWP-769](https://consensyssoftware.atlassian.net/browse/MCWP-769)

## Checklist

### DIFF
--- a/packages/connect-multichain/src/multichain/transports/mwp/index.ts
+++ b/packages/connect-multichain/src/multichain/transports/mwp/index.ts
@@ -457,13 +457,22 @@ export class MWPTransport implements ExtendedTransport {
 
     const cleanup = () => this.dappClient.off('connected', runOnResumeHandler);
 
-    return Promise.race([
-      resumeDeferred.promise,
-      timeoutDeferred.promise,
-    ]).finally(() => {
-      clearTimeout(timeout);
-      cleanup();
-    });
+    return Promise.race([resumeDeferred.promise, timeoutDeferred.promise])
+      .catch(async (error) => {
+        // Defense in depth (complements the SessionStore wipe in
+        // `disconnect()`): the wallet may have already revoked this pairing
+        // out-of-band, so resuming it fails/times out here instead. Delete
+        // it from SessionStore so a subsequent connect() doesn't keep
+        // retrying the same dead pairing.
+        if (session.id) {
+          await this.#deleteMwpSession(session.id);
+        }
+        throw error;
+      })
+      .finally(() => {
+        clearTimeout(timeout);
+        cleanup();
+      });
   }
 
   /**
@@ -748,7 +757,18 @@ export class MWPTransport implements ExtendedTransport {
         this.windowFocusHandler = undefined;
       }
 
-      await this.dappClient.disconnect();
+      // Ensure a later connect() can't pick up a dead handshake URI from a
+      // pending connection attempt that never completed.
+      await this.removeStoredPendingSessionRequest();
+
+      // Wipe the MWP pairing (relay channel + ECIES keys) from SessionStore
+      // regardless of in-memory `dappClient.session` state. `dappClient.disconnect()`
+      // alone is a no-op when the in-memory session is already null (e.g. after a
+      // failed resume), which leaves the SessionStore entry behind. The next
+      // connect() reads that leftover entry via getActiveSession() and tries to
+      // resume a pairing the wallet has already revoked, which times out.
+      // See: https://consensyssoftware.atlassian.net/browse/MCWP-822
+      await this.#clearMwpPairing();
     }
 
     this.notifyCallbacks({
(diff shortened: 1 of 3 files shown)


tommasini:
## Real validation of the fix (real `SessionStore`, not mocked)

In addition to the unit tests added in this PR (which mock `SessionStore` via `vi.spyOn`), I wrote a throwaway script that exercises `MWPTransport` against the **real, installed `@metamask/mobile-wallet-protocol-core` `SessionStore` class** (no mocking of the storage layer at all). Only the `DappClient` is faked, and the fake faithfully reproduces the documented, no-op-when-null behavior of the real `BaseClient.disconnect()`:

```ts
// @metamask/mobile-wallet-protocol-core BaseClient (external dependency)
public async disconnect(): Promise<void> {
  if (!this.session) return; // <-- root cause: SessionStore keys survive
  ...
}
```

**Scenario:** seed a real pairing in `SessionStore`, set the fake `dappClient.session = null` (simulating the exact precondition from the ticket — a prior failed `resume()` already nulled the in-memory session), then call `transport.disconnect()` followed by `transport.connect()`, and observe what actually happens to the persisted pairing and which code path `connect()` takes.

I ran this scenario against **both** the pre-fix code (`98cb7ee`, this PR's parent commit) and the fix in this PR, with the tree restored and `git diff` clean afterward (no test artifacts were committed).

### ❌ Before this fix (`98cb7ee`) — bug reproduces exactly as described in MCWP-822

```
[setup] Seeded SessionStore with a pairing: pairing-mcwp-822-demo
[setup] SessionStore.list() -> [ 'pairing-mcwp-822-demo' ]

[precondition] dappClient.session = null (simulating a prior failed resume())
[precondition] transport.getActiveSession() -> pairing-mcwp-822-demo

[step] Calling transport.disconnect() ...
[step] dappClient.disconnect() was called 1 time(s)

[RESULT] SessionStore.list() after disconnect() -> [ 'pairing-mcwp-822-demo' ]
[RESULT] transport.getActiveSession() after disconnect() -> { id: 'pairing-mcwp-822-demo', channel: 'demo-channel', ... }
[RESULT] ❌ BUG REPRODUCED -- pairing survived disconnect() despite dappClient.session being null.

[step] Calling transport.connect() again ...
[step] transport.connect() REJECTED: SESSION_NOT_FOUND: wallet has already revoked pairing "pairing-mcwp-822-demo"

[RESULT] dappClient.resume() call count: 1
[RESULT] dappClient.connect() call count: 0
[RESULT] ❌ connect() took the RESUME path against a pairing the wallet already revoked (the stuck/timeout scenario from the ticket).

[final] SessionStore.list() -> [ 'pairing-mcwp-822-demo' ]
```

`disconnect()` left the dead pairing in `SessionStore`, so the very next `connect()` tried to resume it, got rejected (in real life this eventually times out instead), and the dapp is stuck — exactly the bug reported in the ticket, and it would keep repeating on every subsequent `connect()` attempt (including after a page refresh, since `SessionStore` is backed by persistent storage).

### ✅ After this fix — pairing is wiped, next `connect()` starts fresh

```
[setup] Seeded SessionStore with a pairing: pairing-mcwp-822-demo
[setup] SessionStore.list() -> [ 'pairing-mcwp-822-demo' ]

[precondition] dappClient.session = null (simulating a prior failed resume())
[precondition] transport.getActiveSession() -> pairing-mcwp-822-demo

[step] Calling transport.disconnect() ...
[step] dappClient.disconnect() was called 1 time(s)

[RESULT] SessionStore.list() after disconnect() -> []
[RESULT] transport.getActiveSession() after disconnect() -> undefined
[RESULT] ✅ Pairing was wiped -- getActiveSession() correctly resolves undefined.

[step] Calling transport.connect() again ...
[step] transport.connect() resolved successfully.

[RESULT] dappClient.resume() call count: 0
[RESULT] dappClient.connect() call count: 1
[RESULT] ✅ connect() took the START path (#startSession) -- no attempt to resume a dead pairing.

[final] SessionStore.list() -> []
```

Same precondition, same fake wallet behavior — but now `disconnect()` unconditionally wipes the pairing from `SessionStore` (independent of `dappClient`'s in-memory session state), so `getActiveSession()` correctly returns `undefined`, and the subsequent `connect()` goes straight to `#startSession` instead of retrying a pairing the wallet already revoked. No more stuck/timeout state.

### Methodology notes

- Storage layer (`SessionStore`) is the **real** class from the installed `@metamask/mobile-wallet-protocol-core` dependency — not mocked or spied.
- Only `DappClient` is faked, and only to deterministically reproduce the documented `disconnect()`/`resume()` behavior from the external package (no real relay/device connection is available in this environment).
- Verified with `git diff` that swapping in the pre-fix file and back left the working tree byte-identical to `HEAD` — this was a temporary, non-committed repro, not a code change.
- Full permanent test suite (`packages/connect-multichain/src/multichain/transports/mwp/index.test.ts`, 29 tests) still passes against the restored fixed code.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -457,13 +457,22 @@ export class MWPTransport implements ExtendedTransport {
...
+    return Promise.race([resumeDeferred.promise, timeoutDeferred.promise])
+      .catch(async (error) => {
+        // Defense in depth (complements the SessionStore wipe in
+        // `disconnect()`): the wallet may have already revoked this pairing
+        // out-of-band, so resuming it fails/times out here instead. Delete
+        // it from SessionStore so a subsequent connect() doesn't keep
+        // retrying the same dead pairing.
+        if (session.id) {
+          await this.#deleteMwpSession(session.id);
+        }
```
adonesky1, line 469: I think this will force a fresh QR/deeplink on any resume failure, not just a dead pairing. connect() already wipes SessionStore via `dappClient.disconnect()` when the in-memory session is still set, including the 10s timeout path. The hole is only `resume()` nulling that session without deleting storage, so move this delete into the `dappClient.resume().catch`.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -759,6 +779,65 @@ export class MWPTransport implements ExtendedTransport {
...
+    const sessionStore = await SessionStore.create(this.kvstore);
+
+    try {
+      const sessions = await sessionStore.list();
+      await Promise.all(
+        sessions.map(async (session) => sessionStore.delete(session.id)),
+      );
+    } catch (error) {
+      logger('error clearing MWP pairing from SessionStore', error);
+    }
```
adonesky1, line 815: I don't think we want to swallow list/delete failures here. If `dappClient.session` is already `null`, the `disconnect()` below is a no-op, so we return success while the pairing is still in SessionStore.

packages/connect-multichain/src/multichain/transports/mwp/index.ts
```diff
@@ -759,6 +779,65 @@ export class MWPTransport implements ExtendedTransport {
...
+  /**
+   * Deletes a single pairing entry from `SessionStore` by id. Used as
+   * defense-in-depth when resuming a specific session fails, so a
+   * subsequent `connect()` doesn't try to resume the same dead pairing
+   * again (see `#clearMwpPairing` for the full wipe done on disconnect).
+   *
+   * @param sessionId - The id of the `SessionStore` entry to delete.
+   * @returns Nothing
+   */
+  async #deleteMwpSession(sessionId: string): Promise<void> {
```
jiexi: Multiple sessions aren't actually supported on the dapp side. I wonder if it makes sense to just use #clearMwpPairing instead which deletes all sessions just to be safe

